PR #2, opened 2015-08-04
Worker

ghost:
This is what I have for the worker. It allows users to specify multiple tasks in their .taskclusterrc file, and also only run tasks on particular kinds of events (pull request, push, or both). I haven't added decent testing yet; but would like to go ahead and get this code out. 

I have deployed this code to: http://experiments.mrrrgn.com where around 10 projects are making use of it to good effect. For example: https://github.com/mozilla/webmaker-android/pull/2425

@jonasfj @gregarndt ; let me know if this is crazy bad. @gregarndt I left the imports as var lib = require('lib'); for now for consistency; but can change that if need be.

@djmitche tagging you since you were on the server review :)


djmitche:
:thumbsup: from me - I like the split between the webhook service and the worker

bin/worker.js
```diff
@@ -0,0 +1,139 @@
...
+var base              = require('taskcluster-base');
+var Promise           = require('promise');
+var exchanges         = require('../lib/exchanges');
+var worker            = require('../lib/worker');
+var _                 = require('lodash');
+var taskcluster       = require('taskcluster-client');
+var GitHubAPI         = require('github');
+
+/** Launch worker */
+var launch = async function(profile) {
```
gregarndt, line 15: Perhaps in the future this could be refactored to provide a common runtime/context for both the worker and server process so the setting up of the config, statsdrain, influx, etc are in one location for both.

lib/worker.js
```diff
@@ -0,0 +1,61 @@
...
+var slugid      = require('slugid');
+
+var worker = module.exports = {};
+
+/**
+ * If a .taskclusterrc exists, attempt to turn it into a taskcluster
+ * graph config, and submit it to the scheduler.
+ **/
+worker.webHookHandler = async function(message, context) {
+  debug('handling pull request: ', message);
```
gregarndt: could this be handling push messages as well?

bin/worker.js
```diff
@@ -0,0 +1,139 @@
...
+
+  // Start monitoring the process
+  base.stats.startProcessUsageReporting({
+    drain:      statsDrain,
+    component:  cfg.get('taskclusterGithub:statsComponent'),
+    process:    'worker'
+  });
+
+  let pulseCredentials = cfg.get('pulse')
+  if (pulseCredentials.username && pulseCredentials.password) {
```
gregarndt: Two things I've seen in some of the taskcluster related things is either using "assert" to verify something is in the config if it would result in throwing an exception:

``` js
assert(pulseCredentails.username, 'Username must be supplied for Pulse connection')
```

or 
instead of wrapping the creation of the pulselistener, throw early...

``` js
if (!pulseCredentials.username || !pulseCredentials.password) throw new Error('Missing pulse credentails');
```

Either way, nothing critical

lib/worker.js
```diff
@@ -0,0 +1,61 @@
...
+  let msgPayload = message.payload;
+  // Try to fetch a taskclusterrc file for every request
+  let taskclusterrc = await request(tcrc.buildConfigUrl(msgPayload));
+  if (taskclusterrc) {
+    try {
+      let scheduler = new taskcluster.Scheduler(context.cfg.get('taskcluster'));
```
gregarndt: Any value of adding a scheduler client to the context instead of instantiating a new one on each pulse message?  I ask because I'm genuinely curious for the things I do.

lib/worker.js
```diff
@@ -0,0 +1,61 @@
...
+          state:       'error',
+          description: e.toString()});
+    }
+  };
+};
+
+/**
+ * Post updates to GitHub, when the status of a task changes.
+ **/
+worker.graphStateChangeHandler = async function(message, context) {
```
gregarndt, line 68: Love this.  Lot cleaner and more informative than just the plain comments that get added in other projects.

lib/worker.js
```diff
@@ -0,0 +1,61 @@
...
+      state:        github.StatusMap[message.payload.status.state],
+      target_url:   inspectorUrl + message.payload.status.taskGraphId,
+      description:  'TaskGraph: ' + message.payload.status.state,
+      context:      'TaskCluster'});
+    // A resulting object indicates failure
+    if (statusUpdate) {
+      throw Error(statusUpdate);
+    }
+  } catch(e) {
+    debug(e);
```
gregarndt: Any value into prefixing this with something about what it was trying to do?

gregarndt:
Looking good, this is going to be great to use.

bin/worker.js
ghost, line 15: Agreed, this is super redundant. I'll get to fixing it up. :)

lib/worker.js
ghost: It does, I just goof'd the debug message. _doh_ :)

lib/worker.js
ghost: Ah, yikes, I think this was just me getting sloppy. I'll move it into the context. Thanks! :)

lib/worker.js
ghost: For sure, good call. :)

lib/taskclusterrc.js
```diff
@@ -0,0 +1,174 @@
...
+function buildTaskSkeleton() {
+  let defaultCreated = new Date();
+  let defaultDeadline = new Date(defaultCreated);
+  defaultDeadline.setHours(defaultDeadline.getHours() + 24);
+  return {
+    provisionerId: 'aws-provisioner-v1',
+    workerType: 'b2gtest',
+    retries: 5,
+    metadata: {
+     owner: 'taskcluster-github@mozilla.com',
```
jonasfj: Owner is the person who created the PR or the push.
Github won't supply emails for all users, but a lot of user have a valid email we should supply that...
If there is no valid email, we should use same pattern as github: `<username>@users.noreply.github.com`

See: https://help.github.com/articles/keeping-your-email-address-private/

Note: whatever string must be a valid email format..
JSON schema will validate this! Later we may use this to allow persona authed users to cancel/rerun their own tasks.. Or we use it for stats... Either way we should strive to have something that identifies the person. And if not a real email, then a consistent pattern used elsewhere too.

lib/taskclusterrc.js
```diff
@@ -0,0 +1,174 @@
...
+     name: 'taskcluster-github graph',
+     description: 'A task graph submitted by the taskcluster-github worker',
+     source: 'https://github.com/taskcluster/taskcluster-github'
```
jonasfj: URL of the `taskclusterrc` file from source repo. Ideally a URL people can open a use to find a blame button... This is to find people who maintain the tajs definition.

ghost: Right now a link to the file is added during the merge of the skeleton->github->user configs: https://tools.taskcluster.net/task-graph-inspector/#AZPuc2tFR9WQhJzb8jCc4g/ 

One issue is that I'm only supporting .taskclusterrc files being used from the base repository (otherwise a random person can assign themselves extra scopes, etc....) That being the case, maybe I should just link back to the pull request here.

djmitche:
I would think the maximum scopes would be specified _outside_ of .taskclusterrc, and anything up to that maximum would be available from any .taskclusterrc, whether in a PR's head or in master.

The "outside" might be on a per-org or per-repo basis.

gregarndt:
@djmitche do you mean something like mozilla-taskcluster does with default scopes for a graph? Not that I'm suggesting it be implemented the same way, just was trying to understand possible solutions.

djmitche:
Yes, similar to that, but configured via some kind of API.

lib/taskclusterrc.js
```diff
@@ -0,0 +1,209 @@
...
+ **/
+taskclusterrc.processConfig = function(taskclusterrc, payload, headUserInfo) {
+  return new Promise(function(accept, reject) {
+    try {
+      let taskclusterConfig = yaml.load(taskclusterrc);
+      if (!taskclusterConfig.tasks) {
+        // fall back to assuming that the user only wanted to configure a single
+        // task when no top level "tasks" key exists in the config.
+        taskclusterConfig.tasks = [taskclusterConfig];
+      }
```
jonasfj, line 46: I don't like this, isn't it saner to ask users to be explicit.
IMO, we need this to be a format that we can support for a long time.

jonasfj:
So I think we need to think about the `.taskclusterrc` format.
At a very minimum we need a version number, maybe even some top-level properties.

Example suggestion:

``` yaml
version: 0
tasks:
- retries: 5
  payload:
    image: "..."
- retries: 5
  payload:
    image: "..."
```

Note, apart from `version` we might also need other top-level options... like task-graph description and other task-graph specific properties, though most of these should probably be auto-generated.

I could also imagine we would want top-level properties for activation of additional `taskcluster-github` features, like bugzilla integration, auto-landing, perhaps even treeherder integration.

I recommend that we proceed using `version: 0` and break compatibility without warning... then we can allow  `version: 1` when we have a stable format and promise to not break compatibility.

We should absolutely also publish a JSON schema for `.taskclusterrc`, and validate against it.
This is simple using `base.validator` which you already load for API.

Also we need to discuss what to do about defaults. Hardcoded defaults can be specifying in JSON schema and `base.validator` will substitute in those values if properties aren't specified. But this doesn't work for `created`, `deadline`, `expires`, etc...

Ideally, I suggest that we don't make any defaults and introduce a special YAML syntax like:
`!from-now '1 day 2 hours'` that constructs a JSON date strong using `taskcluster.fromNowJSON()`.
It'll be a bit of work copying this around, but things would be rather consistent.

If we want some default, ie. a generic worker type, and we want that to be specified from server, so that we don't have to update a million `.taskclusterrc` files when we rename a workerType we could specify it with something like `!const 'default-docker-worker'`.

I'm not sure what is sane to do for defaults. I'm open to suggestions here. But a version number seems sane to add.

lib/taskclusterrc.js
ghost, line 46: Great point; agreed. I'll use the top level version: # format you suggested to decide how the config will be parsed and add a schema for .taskclusterrc. Thanks for these comments/reviews !

ghost:
</img> Added schema validation which will warn github users of any errors encountered and link them back to the schema itself. Starting the schema at v1 so that it matches the app version (i.e. github/v1/_) *note_ this hasn't been deployed in production, so testing users have no expectation of consistency despite version numbers.

Moving toward making defaults explicit in the config file now. That is: forcing the user to declare when they want a field set by us server side.

ghost:
Looking at using [swig](http://paularmstrong.github.io/swig/docs/extending/#tags) for merging in defaults that need to be calculated/added from GitHub meta data. The time functions could be easily implemented with a custom tag.

jonasfj:
hmm.. I just realized that if we use `swig` or custom YAML tags for time functions and defaults, then we can't version the templating logic. Because that happens before we can read the version number.

I previously toyed with idea of structured templating, ie. templating after parsing. I wrote [json-parameterization](https://www.npmjs.com/package/json-parameterization) for this.  It's not very flexible, it only does substitutions in strings. But the concept and the way it's built might be interesting. 

Of course this easily leads to something like cloud formation, where you do something like:

``` yaml
deadline: {$method: from-now, arguments: ['2 days 3 minutes']}
payload: ...
```

That's probably too enterprisey, but a pattern like `{$<method-name>: <argument>}` could perhaps work. Example:

``` yaml
deadline: {$from-now: 2 days 3 minutes}
expires: {$if: [is-pr, {$from-now: 30d}, {$from-now: 1 year}]}
payload:
  workerType: {$const: default-docker-worker}
metadata:
  description: {$template: "Task created for {{revision}} by {{email}}"}
```

Not sure how powerful it has to be... Insertion of constants and `from-now` seems like the most important thing.
If it's not clear I argue this because we can read the version number, before we decide what logic to use to substitute in constants. IMO, this is one of those places where maintaining backward compatibility makes sense, because it's a pain to update old repositories. Implementation wise it's probably just another abstraction.

jonasfj:
Btw, I love that the schema errors shows up on github. It would be a pain to debug otherwise :)

ghost:
Planning to add a few more test cases and merge this today. Will add integration tests via another issue.

PR #18, opened 2015-09-09
Refactor 1

ghost:
After this conversation with @jonasfj https://pastebin.mozilla.org/8845613 I'm making the following changes here:
- simplifying the config schema and setting its version to 0
- renaming .taskclusterrc to .taskcluster.yml
- removing all default values other than created/deadline
- namespace template variable names with dot notation (e.g. github.head.user.email instead of headUserEmail).
- making github provided environment variables optional
- getting rid of the user/org whitelist and deferring security to github (users who can push may trigger taskcluster jobs as well. If that's too restrictive I may just allow all users in the wider org trigger jobs too since that's almost certainly secure enough)


ghost:
@jonasfj I believe this moves us much farther in the direction we'd discussed late last week. Here's an example of the new style `.taskcluster.yml` that this patch enables: https://github.com/ninethings/website/blob/master/.taskcluster.yml

lib/taskcluster-config.js
```diff
@@ -0,0 +1,148 @@
...
+    'scheduler:*'
+  ];
+
+  taskclusterConfig.routes = [
+    `taskcluster-github.${ payload.organization }.${ payload.repository }.` + payload.details['event.head.sha']
+  ];
+
+  // each task can optionally decide if it wants github specific environment
+  // variables added to it
+  taskclusterConfig.tasks = taskclusterConfig.tasks.map((task) => {
```
ghost, line 60: @jonasfj The nice thing about just setting `extra.github.env` to true is that it allows me to merge github environment variables in with whatever others the user has specified. Trying to do that with parameterization alone is a real bummer, even aside from the need to modify the library.

lib/taskcluster-config.js
```diff
@@ -0,0 +1,148 @@
...
+  for (var i in expressions ) {
+    let exp = RegExp(expressions[i], 'i');
+    // just join values so that we don't have to compare multiple times
+    if (exp.test(values.join(' '))) return true;
+  }
+  return false;
+};
+
+function genGitHubEnvs(payload) {
+  return {
```
ghost, line 25: This dictionary could be auto-generated from the details.

lib/taskcluster-config.js
jonasfj, line 60: He he, I understand why you go this way... it's very pragmatic, but not very elegant :)

Hmm... putting my head into thinking mode for few secs...

jonasfj, line 60: Okay, I just did object substitution and proposed a more expressive format on:
https://github.com/jonasfj/json-parameterization/pull/1

I know I mixed the discussion a bit... but just started writing at this is where it ended..

lib/taskcluster-config.js
jonasfj, line 25: it's nice to be explicit... :)

jonasfj:
Looks like a good improvement to me...

ghost:
Thanks !

PR #22, opened 2015-09-14
Add python tox and call apt-get clean

rail:


rail:
Hmm, maybe python-tox is redundant here... It can be installed runtime...

PR #32, opened 2015-09-18
Public perms

ghost:
@Callek @jonasfj 

Callek pointed out that TCGH wouldn't be useful to him without the ability to have fully public tasks. This doesn't fit well with our github team based permissions, so I've opted to add a flag instead.

`extra:  github: insecure: true` bypasses collaborator checks for a given task. We could also use this to ensure that no secrets scope is added to the task.

Thoughts?


dustin:
I don't know what this is.  What's it have to do with me?

Callek:
I was basically commenting that the inability to have random-outsiders submit a pull request that starts a task as a bad thing (specifically in regards to us needing or not needing to do manual work to make sure the PR actually starts a task).

So that inability would block my own desire to try and use githubtaskcluster for relengapi.

I should note however, that encrypted secrets, are also not necessarily a bad thing, since even travis allows that as well (which are tied to the specific repo location, not the yaml in the repo). Such that if we specified insecure:true, and there was a PR, that could allow the use of said secrets. I'm less concerned on this front, but I can imagine some cases where insecure:true is desired even with secrets.

ghost:
@dustin apologies, I think tagging you was out of habit. Was just looking for feedback on whether or not the "insecure" flag was a dumb idea.

ghost:
Killing this implementation, I like some things I did here; but want to make some things cleaner.

jonasfj:
@mrrrgn,
I think we need to not put tc-gh related things in `task.extra`.. well, at least not properties that `taskcluster-github` uses when it first creates the task...

But yeah, maybe we do need to do something other than exclusively rely on github teams. Like defining which teams have access. Example:

``` yaml
version: 0
tasks:
  - triggers:
      - pull-request-opened
      - pull-request-synchronize
      - pull-request-reopened
      - push
    users:
      - user: jonasfj # any github user called `jonasfj`
      - team: mozilla/snappy # any member of "snappy" team of "mozilla" org
      - organization: mozilla # any member of "mozilla" org
      - repository: mozilla/telemetry # any member of repository "telemetry" from org "mozilla"
      - anyone # anyone can trigger this task...
    definition:
      provisionerId: "..."
      workerType: "..."
      payload: "..."
      ...
```

So that these things aren't specified in `task.extra` as they don't have to be part of the task...

Note. I'm not sure this is an idea design either... Perhaps it's better that we only have a top-level trigger/users keys... Or perhaps triggers and users needs to be merged... that it says "trigger on `pull-request-synchronize` if user is `{team: 'mozilla/snappy'}`"

jonasfj:
note: I do see the use case... this is not feature complete yet...

But I agree that the insecureproposal might not be elegant...

Note, perhaps we need to control which scopes various users have rather than this... in some top-level key (not sure --- maybe I'm just rambling)...

ghost:
I like the idea of having a users field outside of extra. In an earlier version I'd implemented this with a whitelist (same idea), but it lived in the extra field and was a bit sloppy.

I'm playing around with the idea of assigning scopes. This could work well with the secrets implementation we've been talking about.

jonasfj:
@mrrrgn,
Please outline a proposal for how the taskcluster.yml will look and use this before we implement it...
I think using scopes somehow (not sure how) could be cool...

jonasfj:
@mrrrgn, random note... I ran into: http://www.emoji-cheat-sheet.com/
If we implement the comment tags we talked about we use emoji...

Like `:unlock:` :unlock: for "allow PR from untrusted user to run",
and  `:+1:` :+1: for "PR approved, auto-land PR" (or maybe `:rocket:` :rocket: is more cool).

I'm sad that github doesn't have an emoji for `:r+:`, but `:+1:` is likely close enough..

petemoore:
@mrrrgn i think @dustin was confused because you meant @djmitche :)

djmitche:
/me wave to Dustin.. this is not the first time :)

AIUI the request is to have the ability for `.taskcluster.yml` (the reliable one on the master branch) optionally build anyone's pull request, not just those on the team or a whitelist.  As such I think Jonas's suggestion of Sep 20 is the right choice -- make a more flexible whitelist, and include an "anyone" option.

I think the idea of using scopes would be to use a role to execute the task, where the role name is based on the creator of the PR (or the unlocking commenter).  That would end up in a spaghetti of github-related roles very quickly, and probably be more than most casual tcgh users would ever want to understand (meaning we'd have to set it up and maintain it for most people - a big barrier).

imbstack:
Closing for now since this seems to lost to the sands of time. Reopen if it should still be worked on in this PR.

petemoore:
@imbstack I think this went stale mostly because @mrrrgn has moved teams, but if @Callek still needs it, maybe @dividehex could take a look as module owner?

PR #37, opened 2015-11-02
Bug 1218523: use roles to define task graph scopes

djmitche:


jonasfj:
Looks nice to me... leaving it to @mrrrgn  to roll this... I think she knows the customers.

ghost:
:+1:

PR #38, opened 2015-12-16
Bug 1232963: bumped taskcluster-base version

petemoore:
Bumped taskcluster-base version to at least same version used in taskcluster-secrets to see if this fixes bug 1232963.

Completely untested ;)

Hoping this will trigger some unit tests, to see if I totally broke things or not... :p

### DIFF
--- a/bin/server.js
+++ b/bin/server.js
@@ -64,7 +64,6 @@ var launch = async function(profile, publisher) {
     context:          {publisher, cfg, githubAPI},
     validator:        validator,
     authBaseUrl:      cfg.get('taskcluster:authBaseUrl'),
-    credentials:      cfg.get('taskcluster:credentials'),
     publish:          cfg.get('taskclusterGithub:publishMetaData') === 'true',
     baseUrl:          cfg.get('server:publicUrl') + '/v1',
     referencePrefix:  'github/v1/api.json',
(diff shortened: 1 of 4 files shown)


petemoore:
@jonasfj I'm not quite sure how to fix this. We're getting http 401 in all the tests - I'm guessing the mock auth server behaves differently in newer taskcluster-base releases.

jonasfj:
hmm... I'm not sure... we probably have to run it locally.
Could very well be this:
https://github.com/taskcluster/taskcluster-github/blob/bug1232963/bin/server.js#L67

Supplying `credentials:      cfg.get('taskcluster:credentials'),` means it'll use the old auth method... which won't work anymore. I think that could be the issue..

Try removing that line.

jonasfj:
See:
https://github.com/taskcluster/taskcluster-lib-api/blob/26aa059a7364a58896364e26d3fff00536f1694f/src/api.js#L1444-L1448

Docs are pretty clear on this... you need to remove credentials to use the remote auth thing...

jonasfj:
Not that I'm parfectly sure that'll fix the issue... sorry :)

petemoore:
@jonasfj yay, that did the trick. thanks for the tip!

petemoore:
I'm not even 100% sure this will get stability added, but I hope so.

petemoore:
@jonasfj could you review again? Thanks!

jonasfj:
Looks good to me... `r+`...

Maybe, jake (:dividehex) wants to merge and land it.

petemoore:
@dividehex one for you! :)

PR #40, opened 2016-03-02
Allow Tasks to be filtered by base repo branch.

bhearsum:
I've been working on using Taskcluster to push production-ready Docker images to Dockerhub. The way this works is that upon pushes to https://github.com/mozilla/balrog, we built and push images. In order to get the secrets needed to do so, I need to give that task the "secrets:get:repo:github.com/mozilla/balrog:dockerhub" scope. This works fine, except that scope is only available to repo:github.com/mozilla/balrog:branch:master, because we don't want pull requests or other potential branches in that repo to build production images.

This patch adds the ability to filter tasks on the base repo's branch, which supports my use case and I think is generally pretty useful.

I'm not a JS dev by any means, but this appears to work - my second test failed prior to my change to taskcluster-config.js.

### DIFF
--- a/lib/taskcluster-config.js
+++ b/lib/taskcluster-config.js
@@ -113,7 +113,9 @@ taskclusterConfig.processConfig = function(params) {
         // Filter out tasks that aren't associated with the current event
         // being handled
         let events = taskConfig.task.extra.github.events;
+        let branches = taskConfig.task.extra.github.branches;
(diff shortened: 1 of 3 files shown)


dividehex:
This looks fine to me.  Docs will also need to be updated to keep in sync with the added functionality.
https://github.com/taskcluster/taskcluster-docs/blob/master/services/taskcluster-github/index.md

Thanks for the PR!

PR #42, opened 2016-03-10
Set correct repo scope for pushes.

bhearsum:
I thought this was happening already, but apparently it's not! Currently, we grant assume:repo:...:pull-request even for pushes.


petemoore:
Thanks for picking this up Ben! :+1:

PR #44, opened 2016-04-22
node 5 + shrinkwrap + new tc-base + eslint + babel compile

imbstack:
###### Other note: The tests are only passing locally right now. I need to figure out what to do w.r.t. the docker image this gets tested in in taskcluster. Let's not merge this until that's figured out.

While I set about working on [bug 1259203](https://bugzilla.mozilla.org/show_bug.cgi?id=1259203) I figured I would get around to modernizing this as much as was convenient for now. It's a big diff, but most changes are fairly mechanical and similar to what I've done to our other projects recently.  I'm not convinced this will work on deploy the first time, so if we have a set of manual testing procedures for deploy, I'd like to know them (or figure them out if we don't have them).

npm-check is happy with us for once:

```
❤️  Your modules look amazing. Keep up the great work. ❤️
```

@dividehex, I'm happy to take care of this project more long-term if you'd like as well. I'm happy either way.

@jonasfj: perhaps you can look at this as well, since you've reviewed most of my other changes like this one.


imbstack:
ok, this is working now!

PR #49, opened 2016-05-04
Stop commenting on every push

imbstack:
This makes tc-github a bit more verbose than we would like. If we'd like, we can go ahead and make this configurable rather than deleting it.

Removing this will make it so that it will be hard to find a graph for a given push you make, but if the task is indexed, you can find it there no problem.

### DIFF
--- a/src/worker.js
+++ b/src/worker.js
@@ -86,15 +86,6 @@ worker.webHookHandler = async function(message, context) {
     });
     if (graphConfig.tasks.length) {
       let graph = await context.scheduler.createTaskGraph(slugid.nice(), graphConfig);
-      // On pushes, leave a comment on the commit
-      if (message.payload.details['event.type'] == 'push') {
-        await github.addCommitComment(
-          context.github,
-          message.payload.organization,
-          message.payload.repository,
-          message.payload.details['event.head.sha'],
-          'TaskCluster-GitHub: ' + INSPECTOR_URL + graph.status.taskGraphId);
-      }
     } else {
       debug('graphConfig compiled with zero tasks: skipping');
     }


jonasfj:
If we already create the statuses:
https://developer.github.com/v3/repos/statuses/#create-a-status

Then yeah, let's get rid of this...

If we don't create the statuses yet, I propose we hold this off until we are doing so...
From what I see on this PR we are creating them...

Standard8:
For pushes to the main branch, how would users see the status? Currently github doesn't seem to show the status of a branch anywhere. So if a direct push to the master (or other branch) failed, I don't think it would be visually reported anywhere at the moment, and you couldn't easily find out the history - afaict taskcluster doesn't support that.

Travis does at least have an interface where you can search for the current status of a branch/repository.

imbstack:
> For pushes to the main branch, how would users see the status? Currently github doesn't seem to show the status of a branch anywhere. So if a direct push to the master (or other branch) failed, I don't think it would be visually reported anywhere at the moment, and you couldn't easily find out the history - afaict taskcluster doesn't support that.

If you're relying on this functionality now, we certainly shouldn't take it away. Perhaps I'll add a `.taskcluster.yml` flag that allows a particular task to opt-out of this email instead. I'd prefer that it's an opt-in, but for backwards compatibility's sake, this seems like the way forward.

> Travis does at least have an interface where you can search for the current status of a branch/repository.

I think it would be nice if Taskcluster did this as well. I think a reasonable first step is to have tc-github automatically add an index route. Something like `github.<project>.<branch>`? Does that make sense to you, @jonasfj? Once that exists it should be fairly straightforward to create a page that lists the status of most recent runs and links into the graph for deeper inspection if necessary. Ultimately that kinda sounds like treeherder, but I guess we can't use that for tc-github projects right now?

jonasfj:
`github.<project>.<branch>.<hash>` might be better... but yeah...

djmitche:
Index doesn't help finding failed jobs

imbstack:
> Index doesn't help finding failed jobs

@jonasfj I think this might be more fuel to our change-how-index-works fire? This would be covered by the changes we discussed in Portland, right? I can't find an etherpad about it. Did you store it in one of your meta-etherpads?

jonasfj:
@imbstack:
https://public.etherpad-mozilla.org/p/jonasfj-bstack-talked-about-index-tweaks

Standard8:
> For pushes to the main branch, how would users see the status? Currently github doesn't seem to show the status of a branch anywhere. So if a direct push to the master (or other branch) failed, I don't think it would be visually reported anywhere at the moment, and you couldn't easily find out the history - afaict taskcluster doesn't support that.

I'm not sure if this changed recently or if I just missed it before, but I'm now seeing status being reported on the commits display, e.g. 

https://github.com/mozilla/example-addon-repo/commits/master

Given that, I'm not sure we do need the commenting any more.

imbstack:
This PR is being superseded by #77. Closing.

(As @Standard8 mentioned, github now shows status on revisions, so we'll stop commenting now)

PR #50, opened 2016-05-19
Check if the user meant to build this with tc-github

imbstack:
This should resolve both [bug 1274077](https://bugzilla.mozilla.org/show_bug.cgi?id=1274077) and [bug 1274322](https://bugzilla.mozilla.org/show_bug.cgi?id=1274322). It's kinda hacky, but I figure it's good enough for now.

### DIFF
--- a/src/worker.js
+++ b/src/worker.js
@@ -29,7 +29,8 @@ worker.webHookHandler = async function(message, context) {
     // Check if this is meant to be built by tc-github at all.
     // This is a bit of a hack, but is needed for bug 1274077 for now
     let c = yaml.safeLoad(taskclusterConfig);
-    if (c.tasks && c.tasks.length > 0 && _.has(c.tasks[0], 'extra.github')) {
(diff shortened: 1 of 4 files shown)


src/worker.js
```diff
@@ -25,6 +26,13 @@ worker.webHookHandler = async function(message, context) {
       ref: message.payload.details['event.base.repo.branch'],
     });
 
+    // Check if this is meant to be built by tc-github at all.
+    // This is a bit of a hack, but is needed for bug 1274077 for now
+    let c = yaml.safeLoad(taskclusterConfig);
+    if (c.tasks && c.tasks.length > 0 && _.has(c.tasks[0], 'extra.github')) {
```
djmitche: I think we should check if _any_ task is a github task.  Eventually, I'd like to put some github tasks in the gecko tree, alongside hg tasks.

imbstack: So perhaps this should be a filter that kicks out tasks that don't have `extra.github` instead?

djmitche: Makes sense -- and if there are none remaining, doesn't proceed.

PR #57, opened 2016-05-23
Slow down errors showing up in sentry

imbstack:
This is the result of me going through the list of errors showing up in sentry for this service right now and trying to stem the tide a bit. I'm pretty certain some of these fixes will help and just guessing about others. If anything seems like a bad/crazy idea, just let me know and I won't do it!

Most of these fixes are for situations where assumptions were made that don't hold, such as the fact that there can be a null `head_commit` and there is an extra action type.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -21,7 +21,7 @@ tasks:
       command:
         - "/bin/bash"
         - "-lc"
-        - "git clone {{event.head.repo.url}} repo && cd repo && git checkout {{event.base.repo.branch}} && npm install . && npm test"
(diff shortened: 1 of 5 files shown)


jonasfj:
Looks sane to me... I don't have a solid understanding of this component... but handling the errors and logging them instead seems okay, when they configuration errors on the github repo...

Perhaps the ideal way to handle configuration errors on a github repo is the use the github status API to report such error...

jonasfj:
r+

imbstack:
Cool. Just as a note, a lot of these aren't the result of misconfiguring on the repo owner's part. It appears to happen all of the time when we push a repo in our org that doesn't have a `.taskcluster.yml` as it stands.

PR #66, opened 2016-07-12
Publish base sha in push pulse messages

gregarndt:
providing the base sha should allow a consumer of this exchange to query github for all commits on a branch since the base sha and iterate over those until head sha is reached.


imbstack:
## r+

Seems reasonable enough!

PR #67, opened 2016-08-02
Add tc-lib-docs

imbstack:
@jonasfj, does this look pretty much how you hoped it would look? I've tested that this works, and I'll push this once we're happy with it!

@kristelteng: Just an FYI this is happening.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -35,6 +35,24 @@ let load = base.loader({
     }),
   },
 
+  docs: {
+    requires: ['cfg', 'validator', 'reference'],
+    setup: ({cfg, validator, reference}) => base.docs({
+      credentials: cfg.taskcluster.credentials,
+      tier: 'core',
+      schemas: validator.schemas,
+      references: [
+        {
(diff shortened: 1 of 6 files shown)


src/main.js
```diff
@@ -35,6 +35,24 @@ let load = base.loader({
     }),
   },
 
+  docs: {
+    requires: ['cfg', 'validator', 'reference'],
+    setup: ({cfg, validator, reference}) => base.docs({
+      credentials: cfg.taskcluster.credentials,
```
jonasfj, line 41: where does it get project name? package.json...? dont we need menu index maybe?

jonasfj:
# `r+`

src/main.js
imbstack, line 41: project name is automatically grabbed from package.json unless otherwise specified.

menuIndex is set to 10 unless otherwise specified. I figure the index will be sorted by
1. menuIndex
2. alphabetically

So that things are stable when we don't care about the ordering.

PR #77, opened 2016-10-14
Port to task-groups

imbstack:
This removes any use of the scheduler in favor of task-groups.

I deployed this for a few minutes just to test and it appears to  work. Examples:
push-inspector: https://tools.taskcluster.net/push-inspector/#/fw6STsCxQ6Cpu1fI7DkNdg/fw6STsCxQ6Cpu1fI7DkNdg?_k=qym78g
PR that used the task-group for testing: https://github.com/TaskClusterRobot/hooks-testing/pull/19

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -1,0 +1,21 @@
+let Entity = require('azure-entities');
+
+module.exports = {};
+
+/**
+ * Entity for tracking which task-groups are associated
+ * with which org/repo/sha, etc.
+ *
+ */
+module.exports.Build = Entity.configure({
+  version: 1,
+  partitionKey: Entity.keys.CompositeKey('organization', 'repository', 'sha'),
+  rowKey: Entity.keys.StringKey('taskGroupId'),
+  properties: {
+    organization: Entity.types.String,
+    repository: Entity.types.String,
+    sha: Entity.types.String,
+    taskGroupId: Entity.types.String,
+    status: Entity.types.String,
+  },
+});
--- a/src/data.js
+++ b/src/data.js
@@ -16,6 +16,6 @@ module.exports.Build = Entity.configure({
     repository: Entity.types.String,
     sha: Entity.types.String,
     taskGroupId: Entity.types.String,
-    status: Entity.types.String,
+    state: Entity.types.String,
   },
 });
--- a/src/data.js
+++ b/src/data.js
@@ -17,5 +17,7 @@ module.exports.Build = Entity.configure({
     sha: Entity.types.String,
     taskGroupId: Entity.types.String,
     state: Entity.types.String,
+    created: Entity.types.Date,
+    updated: Entity.types.Date,
   },
 });
--- a/src/data.js
+++ b/src/data.js
@@ -9,8 +9,8 @@ module.exports = {};
  */
 module.exports.Build = Entity.configure({
   version: 1,
-  partitionKey: Entity.keys.CompositeKey('organization', 'repository', 'sha'),
-  rowKey: Entity.keys.StringKey('taskGroupId'),
+  partitionKey: Entity.keys.StringKey('taskGroupId'),
+  rowKey: Entity.keys.ConstantKey('taskGroupId'),
   properties: {
     organization: Entity.types.String,
     repository: Entity.types.String,
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -6,11 +6,6 @@ let assert = require('assert');
 let _ = require('lodash');
 
 let INSPECTOR_URL = 'https://tools.taskcluster.net/push-inspector/#/';
-let STATUS_MAPPING = {
-  running:  'pending',
-  blocked:  'failure',
-  finished: 'success',
-};
 
 /**
  * Create handlers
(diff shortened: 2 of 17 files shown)


src/handlers.js
```diff
@@ -278,3 +279,14 @@ async function isCollaborator({login, organization, repository, sha, context}) {
...
 }
+
+async function createTasksInGroup(queue, graphConfig) {
+  let taskGroupId = null;
+  let created = [];
+  for (let task of graphConfig.tasks) {
+    taskGroupId = task.task.taskGroupId;
+    created.push(queue.createTask(task.taskId, task.task));
+  }
+  await Promise.all(created);
```
jonasfj: this is a `Promise.all(graphConfig.map(t => queue.createTask(t.taskId, t.task)));`

src/intree.js
```diff
@@ -54,54 +56,57 @@ function completeInTreeConfig(config, payload) {
...
-  }).filter((task) => {
-    // Filter out tasks that aren't associated with the current event
-    // being handled
-    let events = task.task.extra.github.events;
-    let branches = task.task.extra.github.branches;
-    return _.some(events, ev => RegExp(ev).test(payload.details['event.type'])) && (!branches || branches &&
-      _.includes(branches, payload.details['event.base.repo.branch']));
-  });
-  return completeInTreeConfig(config, payload);
+    let taskGroupId = slugid.nice();
```
jonasfj: taskGroupId should be the taskId of the first task from the `.taskcluster.yml` file... The first that isn't filtered out.

We do this because the first task is the decision task (or should be -- in which case there often only one), hence, we get the semantics that `taskGraphId == decisionTask.taskId` which is pretty awesome :)

jonasfj:
This looks a like a good start... but it seems there is a lot to improve..

src/intree.js
```diff
@@ -54,54 +56,57 @@ function completeInTreeConfig(config, payload) {
...
+      return {
+        taskId: slugid.nice(),
+        task: _.extend(task, {taskGroupId}),
+      };
+    }).filter((task) => {
+      // Filter out tasks that aren't associated with the current event
+      // being handled
+      let events = task.task.extra.github.events;
+      let branches = task.task.extra.github.branches;
+      return _.some(events, ev => RegExp(ev).test(payload.details['event.type'])) && (!branches || branches &&
```
jonasfj: For example: `RegExp(ev)` where `ev` is untrusted is actually an attack vector... It's relatively easy to construct regular expressions that takes forever to evaluate, hence, crashing the server -- just some fun to think about :)

Note: not saying that it should be done in this PR. Just saying lots of low hanging fruit in tc-gh.

imbstack: I actually removed this regex in a previous commit, but it turns out that people rely on it being there, despite it being undocumented afaict :/

Perhaps we can define it to be something more safe that still handles what people use already.

imbstack:
@jonasfj I think this is ready now. Things to note:
- This no longer comments on PRs and pushes. The statuses should be enough to keep people in-the-know about what's going on. The statuses exist on both pushes and PRs already. I can add commenting back in if we want, but I find it annoying.

src/data.js
```diff
@@ -0,0 +1,23 @@
...
+ * Entity for tracking which task-groups are associated
+ * with which org/repo/sha, etc.
+ *
+ */
+module.exports.Build = Entity.configure({
+  version: 1,
+  partitionKey: Entity.keys.StringKey('taskGroupId'),
+  rowKey: Entity.keys.ConstantKey('taskGroupId'),
+  properties: {
+    organization: Entity.types.String,
```
imbstack, line 15: @jonasfj should we rename this to `owner` since it can really be a github user _or_ organization?

jonasfj: well, don't we only support organizations?

src/data.js
```diff
@@ -0,0 +1,23 @@
...
+    repository: Entity.types.String,
+    sha: Entity.types.String,
+    taskGroupId: Entity.types.String,
+    state: Entity.types.String,
+    created: Entity.types.Date,
+    updated: Entity.types.Date,
+  },
```
jonasfj, line 22: so these never expires?

src/handlers.js
```diff
@@ -171,23 +190,48 @@ async function jobHandler(message) {
...
+        updated: now,
+      }).catch(async (err) => {
+        if (err.code !== 'EntityAlreadyExists') {
+          throw err;
+        }
+        let build = await this.Builds.load({
+          taskGroupId,
         });
-      }
+        assert.equal(build.state, 'pending', `State for ${organization}/${repository}@${sha}
```
jonasfj, line 232: assert on org, repo, sha, etc?

src/data.js
imbstack: Nope!

src/data.js
imbstack: I don't want to expire them for the time being. We can think about that later, but I'd like them to live for quite a long time.

src/data.js
jonasfj: In that case I guess `owner` is just as good :)

PR #80, opened 2016-10-21
New ping API

owlishDeveloper:
@djmitche @imbstack r?


imbstack (approved):

imbstack:
We need to come up with a way to get you permissions on all of these repos so that the tests can pass :/

I expect all of the diffs are pretty much just this, so I'll go through and merge+deploy on Monday!

djmitche:
> We need to come up with a way to get you permissions on all of these repos so that the tests can pass :/

That sounds like a job for taskcluster-github, to be honest :)

PR #81, opened 2016-10-24
GITHUB_PULL_REQUEST should be a string not a number

petemoore:
GITHUB_PULL_REQUEST env var was appearing as a number in the task.payload.env section, rather than a string (since all env vars are strings). Fixed in this PR.

### DIFF
--- a/src/intree.js
+++ b/src/intree.js
@@ -30,13 +30,14 @@ function completeInTreeConfig(config, payload) {
 
   // each task can optionally decide if it wants github specific environment
   // variables added to it
+  let stringify = x => x ? `${x}` : x;
   config.tasks = config.tasks.map((task) => {
     if (task.task.extra.github.env) {
       task.task.payload.env = _.merge(
         task.task.payload.env || {}, {
           GITHUB_EVENT: payload.details['event.type'],
           GITHUB_BRANCH: payload.details['event.base.repo.branch'],
-          GITHUB_PULL_REQUEST: payload.details['event.pullNumber'],
+          GITHUB_PULL_REQUEST: stringify(payload.details['event.pullNumber']),
           GITHUB_BASE_REPO_URL: payload.details['event.base.repo.url'],
           GITHUB_BASE_USER: payload.details['event.base.user.login'],
           GITHUB_BASE_SHA: payload.details['event.base.sha'],


petemoore:
@gregarndt I'm 99% sure this shouldn't negatively impact docker-worker, but CC'ing just in case there is any special treatment hardcoded in the worker w.r.t. this env variable (which hopefully there isn't).

imbstack:
r+ once we figure out why the tests are failing!

djmitche:
@petemoore any chance you want to finish this?

petemoore:
@djmitche thanks for the reminder - will pull in latest fixes and retrigger tests to see if that resolves it. I'm guessing the failures were not related to the change here.

petemoore:
@djmitche I've no idea why these tests are failing, after merging in the latest changes. Not being a node developer, I'm not sure I'm the best person to discern what the problem might be. As you've been quite active on this project, could you have a quick look to see if you see anything obvious? My change was just to ensure that env vars are created as strings, as required in the payload. Many thanks for any support you can provide.

petemoore:
Note, since the merge 5 days ago of #88 the master branch is also failing, so I took the liberty of also merging in the currently open #89 which was green.

djmitche:
```
  taskcluster-github error (reported to sentry) while calling job handler: TypeError: Cannot read property 'toString' of undefined +22ms
```

djmitche:
So the issue is that `event.pullNumber` isn't always set.  JS has this nice property of propagating `undefined`, so

```
x = {SOME_KEY: some_object['some-property-that-isnt-there'}
```

sets `x` to `{}`.  However, `undefined.to_string()` causes an error.  Of course, `to_string` isn't a method of integers, either.

petemoore:
Thanks #djmitche! That explains it, and nice spot!

> Of course, to_string isn't a method of integers, either.

The change introduced `toString()` rather than `to_string()`. 😜

petemoore:
@djmitche I think the same tests are still failing, also with `stringify` :/

djmitche:
Yeah, this isn't rebased onto #89.  Assuming @imbstack is happy I think we can merge anyway.

imbstack (approved):

petemoore:
Weird - we got [this](https://tools.taskcluster.net/task-inspector/#UkQe0e2OSRaOd1tXShW09Q/) after this PR merged.

petemoore:
@djmitche do you know if taskcluster-github auto-deploys on merge-to-master? I guess I can check heroku...

djmitche:
No, it doesn't auto-deploy.

petemoore:
@djmitche No worries - I checked heroku and saw it was deployed a day later, so I think all is ok again. Thanks for replying though. 👍

PR #83, opened 2016-10-31
Better support scopes and routes

imbstack:


src/intree.js
```diff
@@ -23,10 +24,6 @@ function completeInTreeConfig(config, payload) {
     ];
   }
 
-  config.routes = [
-    `taskcluster-github.${ payload.organization }.${ payload.repository }.` + payload.details['event.head.sha'],
```
jonasfj, line 27: I assume this is intentional...

jonasfj (approved):

src/intree.js
imbstack: Yeah, it doesn't do anything now.

imbstack: We could add it to every task I suppose, but I don't think anyone has noticed it missing.

PR #84, opened 2016-12-22
We now have a list of secrets

owlishDeveloper:
Solution for [the bug "Allow tc-gh to have multiple webhook secrets" ](https://bugzilla.mozilla.org/show_bug.cgi?id=1323025)

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -21,7 +21,7 @@ defaults:
       type: 'oauth'
 
   webhook:
-    secret: !env WEBHOOK_SECRET
+    secret: !env:list WEBHOOK_SECRET
 
   intree:
     provisionerId: !env PROVISIONER_ID
@@ -66,7 +66,7 @@ production:
 
 test:
   webhook:
-    secret: 'abc123'
+    secret: ['abc123', 'def456']
 
   azure:
     account: 'inMemory'
--- a/src/api.js
+++ b/src/api.js
@@ -121,18 +121,20 @@ api.declare({
     return resolve(res, 400, 'Request missing a body');
   }
 
-  let webhookSecret = this.cfg.webhook.secret;
+  let webhookSecrets = this.cfg.webhook.secret;
   let xHubSignature = req.headers['x-hub-signature'];
 
-  if (xHubSignature && !webhookSecret) {
+  if (xHubSignature && !webhookSecrets) {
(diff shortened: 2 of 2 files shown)


owlishDeveloper:
@djmitche @imbstack r?

config.yml
```diff
@@ -21,7 +21,7 @@ defaults:
       type: 'oauth'
 
   webhook:
-    secret: !env WEBHOOK_SECRET
+    secret: !env:list WEBHOOK_SECRETS
```
djmitche: I think we should either make this plural everywhere ("secrets") or singular everywhere ("secret").  The advantage to plural is that it make it clear it's an array.  The advantage to singular is that the existing env variable will work without change.

src/api.js
```diff
@@ -121,18 +121,20 @@ api.declare({
...
-  } else if (webhookSecret && !xHubSignature) {
+  } else if (webhookSecrets && !xHubSignature) {
     return resolve(res, 400, 'Request missing a secret');
-  } else if (webhookSecret && xHubSignature) {
+  } else if (webhookSecrets && xHubSignature) {
     // Verify that our payload is legitimate
-    let calculatedSignature = generateXHubSignature(webhookSecret,
-      JSON.stringify(body));
-    if (!compareSignatures(calculatedSignature, xHubSignature)) {
+    if (!webhookSecrets.some(webhookSecret => {
```
djmitche, line 133: good use of `some` :)

djmitche (changes requested):
As for the errors -- Travis failure is expected (it will always fail since it needs credentials that are only available to the repository in the taskcluster org..).

It looks like the errors in TaskCluster are

    taskcluster-github Server is not setup to handle secrets +17ms

just as you said, due to the env var not being set.  That's handled in `test/fetch_secrets.js` by pulling from the secrets API, and I can fix that up if you'd like.  On the other hand, if you choose to use the singular name (`WEBHOOK_SECRET`) then there's no need to change anything!  Let me know if you'd like me to add that secret.

config.yml
owlishDeveloper: I made it singular, so that the existing vars could work. In the code though the variable that holds the array is still plural.

djmitche (approved):

imbstack (approved):
Awesome.

PR #88, opened 2016-12-29
Bug 1253619 - Added support for release events, try #2

owlishDeveloper:
[The bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1253619)


src/intree.js
```diff
@@ -22,6 +22,11 @@ function completeInTreeConfig(config, payload) {
     config.scopes = [
       prefix + payload.details['event.base.repo.branch'],
     ];
+  } else if (payload.details['event.type'] == 'release') {
+    let prefix = `assume:repo:github.com/${ payload.organization }/${ payload.repository }:branch:`;
+    config.scopes = [
+      prefix + payload.details['event.base.repo.branch'] + ':release',
```
imbstack: I sorta wonder whether or not we care what branch the release came from? I don't really understand releases though, so if this seems right (and if it seems right to bhearsum) this is fine.

imbstack (approved):
🏆

src/intree.js
owlishDeveloper: I think we do (at least Ben does)

PR #89, opened 2016-12-30
Test fixes

djmitche:
This should make the tests a little more reliable.. one hopes.  We'll see how the PR tests go.

### DIFF
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -39,11 +39,13 @@ class Handlers {
     this.statusQueueName = statusQueueName;  // Optional
     this.jobQueueName = jobQueueName;  // Optional
     this.context = context;
+
+    this.handlerComplete = null;
   }
 
   /**
    * Set up the handlers.  If {noConnect: true}, the handlers are not actually
-   * connected to a pulse connection (used for tests)
+   * connected to a pulse connection (used for tests).
    */
   async setup(options) {
     options = options || {};
@@ -83,6 +85,10 @@ class Handlers {
       handler.call(this, message).catch(err => {
         debug(`error (reported to sentry) while calling ${name} handler: ${err}`);
         this.monitor.reportError(err);
+      }).then(() => {
+        if (this.handlerComplete) {
+          this.handlerComplete();
+        }
       });
     };
     this.jobListener.on('message',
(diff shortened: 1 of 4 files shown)


src/handlers.js
```diff
@@ -68,8 +68,16 @@ class Handlers {
     await this.statusListener.bind(queueEvents.taskException({schedulerId: 'taskcluster-github'}));
     await this.statusListener.bind(queueEvents.taskGroupResolved({schedulerId: 'taskcluster-github'}));
 
-    this.jobListener.on('message', this.monitor.timedHandler('joblistener', jobHandler.bind(this)));
-    this.statusListener.on('message', this.monitor.timedHandler('statuslistener', statusHandler.bind(this)));
+    const callHandler = (name, handler) => message => {
+      handler.call(this, message).catch(err => {
+        debug(`error calling ${name} handler: ${err}`);
```
imbstack: Maybe we should say "error reported to sentry" or something along those lines?

test/api_test.js
```diff
@@ -46,55 +46,35 @@ suite('api', () => {
...
     let builds = await helper.github.builds();
-    try {
-      assert.equal(builds.builds.length, 4);
-      assert.equal(builds.builds[0].organization, 'abc123');
-      assert.equal(builds.builds[1].organization, 'ghi789');
-      done();
-    } catch (e) {
-      done(e);
-    }
+    assert.equal(builds.builds.length, 4);
```
imbstack, line 51: iirc, if line 50 `await helper.github.builds();` throws an exception, this test will hang rather than complete now? I might be mis-remembering though.

imbstack (dismissed):

test/api_test.js
djmitche: No, because it's an async function, mocha will wait for the Promise, and an exception in an async function will reject the Promise.

src/handlers.js
djmitche: Good idea

djmitche:
Ok, I'm done playing with this.  Since it was intermittent, I'm not sure (a) which of these fixes stopped the intermittents or (b) if any of them did.  But hopefully fixing async stuff, taking pulse out of the equation, taking queue operations out of the equation, and avoiding polling will be sufficient.

owlishDeveloper (approved):

PR #91, opened 2017-01-04
Corrected mistake in sha retreiving

owlishDeveloper:
I think it might fix https://pastebin.mozilla.org/8958224

@imbstack @djmitche r?

### DIFF
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -146,7 +146,7 @@ async function jobHandler(message) {
     owner: organization,
     repo: repository,
     ref: `refs/tags/${message.payload.details['event.version']}`,
-  });
+  }).sha;
 
   debug(`handling ${message.payload.details['event.type']} webhook for: ${organization}/${repository}@${ref}`);
   let repoconf = undefined;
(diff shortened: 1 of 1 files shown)


src/handlers.js
```diff
@@ -145,8 +145,8 @@ async function jobHandler(message) {
   let sha = message.payload.details['event.head.sha'] || await context.github.repos.getShaOfCommitRef({
     owner: organization,
     repo: repository,
-    ref: `heads/${message.payload.details['event.base.repo.branch']}`,
-  });
+    ref: `refs/tags/${message.payload.details['event.version']}`,
+  }).sha;
```
djmitche, line 179: I assume it's the `.sha` that makes the critical difference in fixing the error, as before the URI included some long JSON blob?  At any rate, this looks good.

djmitche (approved):

djmitche:
I think this should be fine with the fixes #89 applied.. just checking that before I merge.

src/handlers.js
owlishDeveloper: Yeah, I tried this function out separately, and it returns a giant JSON full of goodies. I think this sha gets into request url later, that's why it gets too long.

PR #92, opened 2017-01-04
Another fix for the sha issue

owlishDeveloper:
@imbstack @djmitche r?


djmitche (commented):
This looks right to me, but so did the last patch!  Brian?

imbstack (approved):
I similarly thought the last one would work too! Oh well. This one should be right.

PR #93, opened 2017-01-04
Corrected clone url

owlishDeveloper:
@imbstack I want to include the clone url after all!

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -44,7 +44,7 @@ function getPushDetails(eventData) {
     'event.head.repo.url': eventData.repository.clone_url,
     'event.head.sha': eventData.after,
     'event.head.ref': ref,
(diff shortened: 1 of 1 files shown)


src/api.js
```diff
@@ -44,7 +44,7 @@ function getPushDetails(eventData) {
     'event.head.repo.url': eventData.repository.clone_url,
     'event.head.sha': eventData.after,
     'event.head.ref': ref,
-    'event.base.sha': eventData.before,
+    'event.base.sha': eventData.before,
```
imbstack: I think this is an extraneous space at the end of the line?

imbstack (approved):
Otherwise looks good to me!

src/api.js
owlishDeveloper: Oy vey

PR #97, opened 2017-01-17
Expose repo names (base and head)

petemoore:
For integrating with generic worker tasks, having the repo names available makes coding a lot easier. We could parse the base/head clone URLs, but on a lot of Windows workers, we don't have tools like `sed` build in, so being able to get this data without any string parsing is very useful.


petemoore:
Doc change for this in https://github.com/taskcluster/taskcluster-docs/pull/147.

imbstack (approved):
Looks great. ty.

I'll deploy this shortly.

Issue #99, opened 2017-01-17
Intermittent: some token substitutions not working in a pull request


petemoore:
Compare these two tasks, generated from the same PR:

1) https://queue.taskcluster.net/v1/task/ERiMtxoYSri-M0YWhMzBSA
2) https://queue.taskcluster.net/v1/task/ICyhFSvWTFWYJfmTi15pKA

The tokens in the first task have been properly substituted, such that we have a valid task. The second task still contains the literal text `{{ event.base.user.login }}`.

The missing substitutions seem to be intermittent and happened about one time in six for me (at a guess).

I noticed that `{{ event.base.user.login }}` is a token which is replaced for a pull request, but not replaced for a push, so I wonder if there is a race condition, such that when a commit is pushed to a branch where there is an active pull request, sometimes the push event might win the race against the pull request event, and then the task is generated as if it was for a regular push rather than for a pull request. This is based on looking at the data in https://github.com/taskcluster/taskcluster-github/blob/912b765f7859a083bdb4a9c380f546e5f8407525/src/api.js#L17-L49 and also remembering that when using travis-ci.org, two builds would get triggered on a PR - and joining the dots and suspecting that the reason for this is also that a push to a PR results in both events getting fired.

CC @djmitche @imbstack @owlishDeveloper

djmitche:
Fixed in #100.

PR #100, opened 2017-01-19
Added event.base.ref, event.base.repo.name, event.base.repo.url, event.base.user.login to push metadata

petemoore:
In #99 I was having the problem that `{{ event.base.user.login }}` was intermittently not getting resolved. Although that is a separate issue, it did highlight that we get different data between push events and pull request events.

For the CI task I was running, I needed to `go get` the go package from the github base repository, and then pull changes in from the github head repository (this [peculiar workflow](http://blog.campoy.cat/2014/03/github-and-go-forking-pull-requests-and.html) is an idiosyncrasy of go development).

For Pull Requests, that worked fine, as I had the metadata for both the base repository, and the head repository. However, if the CI task executes as a consequence of a regular push on a branch, rather than a PR, we run into the problem that we don't have a different base and head repository, so we do not have `{{ event.base.user.login }}` for example.

After thinking about this some more, I realised there were essentially two options:

1) to embed logic into the task, so that at runtime (rather than task creation time) the task could detect if it was a pull request or a push that had caused it to be spawned, and then use the _head_ repository settings for the push task, and continue to use both the _base_ and _head_ repositories for the pull request task.
2) to consider that the merge of a pull request is essentially the same as a push, except that the base branch and the head branch happen to be the same branch. In which case, populating both the _head_ and _base_ repository info for a push event with the same data would mean we have a more generic handling.

This is an implementation of option 2, which I consider to be simpler and more intuitive. For the majority of cases, tasks that have been crafted to work with Pull Requests should just work for Push events too. Now a push can be considered to be _like_ a pull request, but where the source and target branches (head and base) are the same.

Note, this could be considered a mild case of sticking a square block through a round hole .... but for the most part I think it will cause broken tests to magically work, and when not, people still have the freedom to craft tasks at runtime based on [`GITHUB_EVENT`](https://docs.taskcluster.net/manual/integrations/github#pull-request-metadata).

This PR therefore adds:

* `event.base.ref`
* `event.base.repo.name`
* `event.base.repo.url`
* `event.base.user.login`

to push events, and sets the value to the same as

* `event.head.ref`
* `event.head.repo.name`
* `event.head.repo.url`
* `event.head.user.login`

At the same time, I also sorted the entries lexicographically, to make it easier to maintain the lists and compare data between different event types.

Lastly, I've also made the data consistent between pull requests events and push events for `{{ event.base.ref }}` and `{{ event.head.ref }}`. Previously, in push events they were refspecs, and in pull request events they were branch names. Now they are consistently a refspec, and the branch names continue to be provided via `{{ event.base.repo.branch }}` and `{{ event.head.repo.branch }}`.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -18,14 +18,14 @@ function sanitizeGitHubField(field) {
 // See https://developer.github.com/v3/activity/events/types/#pullrequestevent
 function getPullRequestDetails(eventData) {
   return {
-    'event.base.ref': eventData.pull_request.base.ref,
+    'event.base.ref': 'refs/heads/' + eventData.pull_request.base.ref,
     'event.base.repo.branch': eventData.pull_request.base.ref,
     'event.base.repo.name': eventData.pull_request.base.repo.name,
     'event.base.repo.url': eventData.pull_request.base.repo.clone_url,
     'event.base.sha': eventData.pull_request.base.sha,
     'event.base.user.login': eventData.pull_request.base.user.login,
 
-    'event.head.ref': eventData.pull_request.head.ref,
+    'event.head.ref': 'refs/heads/' + eventData.pull_request.head.ref,
     'event.head.repo.branch': eventData.pull_request.head.ref,
     'event.head.repo.name': eventData.pull_request.head.repo.name,
     'event.head.repo.url': eventData.pull_request.head.repo.clone_url,
--- a/src/api.js
+++ b/src/api.js
@@ -18,22 +18,22 @@ function sanitizeGitHubField(field) {
 // See https://developer.github.com/v3/activity/events/types/#pullrequestevent
 function getPullRequestDetails(eventData) {
   return {
-    'event.type': 'pull_request.' + eventData.action,
-    'event.pullNumber': eventData.number,
-
-    'event.base.user.login': eventData.pull_request.base.user.login,
+    'event.base.ref': eventData.pull_request.base.ref,
+    'event.base.repo.branch': eventData.pull_request.base.ref,
     'event.base.repo.name': eventData.pull_request.base.repo.name,
     'event.base.repo.url': eventData.pull_request.base.repo.clone_url,
     'event.base.sha': eventData.pull_request.base.sha,
-    'event.base.ref': eventData.pull_request.base.ref,
-    'event.base.repo.branch': eventData.pull_request.base.ref,
+    'event.base.user.login': eventData.pull_request.base.user.login,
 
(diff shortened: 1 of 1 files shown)


petemoore:
Doc changes added in https://github.com/taskcluster/taskcluster-docs/pull/148.

petemoore:
@djmitche Note, since this *adds* properties that were previously not set, this shouldn't break existing tasks. The only exception is that the adapted handling of `{{ event.base.ref }}` and `{{ event.head.ref }}` to make the format consistent across pushes and pull requests.

For this change, we could send out an email to taskcluster-tools and mention that this change is occurring, and that `{{ event.base.repo.branch }}` and `{{ event.head.repo.branch }}` can be used if just the branch name is required.

djmitche:
I'll defer to @owlishDeveloper and @imbstack on this topic.  I think better clarity is good, in general, but I don't know how to judge the impact of this change on our beleaguered users.

owlishDeveloper:
I like that all the keys are organized nicely. That definitely helps!

I followed the documentation link and left my comment there. I'll repeat it here: There is no such thing in the pull-request webhook payload as "refs/heads/whatever". At least in the integration webhook (I didn't bother to compare it with oAuth webhook because tc-gh will be an integration soon anyway, but I do believe that the only difference between these webhooks is installation.id). I didn't even see we can parse anything there to get the "refs/heads/whatever", so it appears to me that the change won't work. (by the way, I just noticed that in the docs right now for push there's 

> GITHUB_HEAD_REF        | "{{ event.head.ref }}"         | bug1394856

this needs to be changed to 

> GITHUB_HEAD_REF        | "{{ event.head.ref }}"         | refs/heads/bug1394856

because that's what's actually there.

owlishDeveloper:
UPD>Looked more carefully :) Yeah, let's add the keys we need. I focused on the word "intermittent", but then I noticed that pushes don't have "event.base.user.login" at all. If that's an inconvenience, let's fix it! 👍

owlishDeveloper (approved):

owlishDeveloper:
While we're at it... I've just been thinking about all those various users - base/head/sender/pusher/and so on - and it appears to me that we're having an inconsistency there.

For example, let's take event.head.user.login. If Brian makes a PR to TaskClusterRobot's repository, in the PR details event.head.user.login will be "TaskClusterRobot", but in the corresponding push - "imbstack". I think this can be confusing and we should either rename the keys that don't align, or align the keys with the same name, or both.

petemoore:
Thanks guys!

PR #102, opened 2017-01-21
Debugging messages to debug release

owlishDeveloper:
Trying to find the reason for intermittent failures of release event handling.

Handlers seem to be nicely equipped with debugging messages, so I focused on server side:

intree.js - check to see if YML is being parsed as expected; catch the taskGroupID for releases.
api.js - check to see if release webhooks arrive and being recognized as such.
exchanges.js - check to see if webhook payload is being parsed.

Also added a bit of clarity so that we instantly could see where the message comes from.


src/api.js
```diff
@@ -171,13 +171,15 @@ api.declare({
     } else if (eventType == 'ping') {
       return resolve(res, 200, 'Received ping event!');
     } else if (eventType == 'release') {
+      debug('Received release event webhook payload. Processing...');
```
imbstack: Maybe add a `// TODO: remove this` here so that we know it's temporary?

src/intree.js
```diff
@@ -114,6 +123,11 @@ module.exports.setup = function(cfg) {
     if (config.tasks.length > 0) {
       let taskGroupId = config.tasks[0].taskId;
       config.tasks = config.tasks.map((task) => {
+        // Temporary for debugging purposes
+        if (task.task.extra.github.events.some(event => event == 'release')) {
```
imbstack: The depth of the dot-access here worries me. Are we sure we won't have an issue with one of these fields not existing or something?

src/intree.js
```diff
@@ -92,6 +92,15 @@ module.exports.setup = function(cfg) {
       'taskcluster.docker.workerType': cfg.intree.workerType,
     }));
 
+    // Temporary loop for debugging purposes
+    config.tasks.forEach(task => {
+      if (task.extra.github.events.some(event => event == 'release')) {
```
imbstack, line 99: Same as below.

imbstack (changes requested):
Looks pretty good! Just some minor nits.

src/intree.js
owlishDeveloper: Well, it comes from [here](https://github.com/taskcluster/taskcluster-github/blob/master/src/intree.js#L101-L104 ) ... I think it should be correct.

owlishDeveloper: A second thought, I can wrap [this whole thing](https://github.com/taskcluster/taskcluster-github/blob/master/src/intree.js#L100-L125) in try-catch block

owlishDeveloper:
Added `// TO DO` comments

src/intree.js
djmitche: It's usually better to be safe than sorry with temporary debugging stuff, so I like the try/catch block.

djmitche (approved):
I'm happy with this once the comments are addressed -- go ahead and push, then we can promote to production.

src/intree.js
djmitche: don't forget this one ^^

owlishDeveloper:
Oh dear... That's messy even by my standards! I'm closing this for now; besides, I think I noticed an error in my code there

src/intree.js
owlishDeveloper: You mean add it to try-catch? I didn't :) However, before messing up commits again, I deployed the thing, and it appears there may be something wrong with parsing YML. I'll look more closely later.

djmitche:
Yeah, it looks like you merged pete's changes, too?

owlishDeveloper:
Yep 😜. It's all because on my local repo, I didn't make a separate branch for this, and I also don't keep my master up-to-date :( What will happen when this PR will get merged - this mess will remain, won't it? Is there a way of repairing this?

djmitche:
A rebase onto an updated master should be able to fix it.  However, if you've already merged things to master, then it may be too late :(

djmitche:
*dustin actually looks at master*

OK, nothing got merged so that's not an issue .
```
git fetch upstream
git rebase -i upstream/master
```
should do the trick -- just drop any commits that aren't about adding debug info (and maybe squash those which are),

owlishDeveloper:
Yeah, I meant my local master (the poor thing just doesn't seem to be able to get a nice quiet life on my computer).

Ok, I'll try that rebase, thank you!

owlishDeveloper:
Yay! It worked, thank you Dustin!!

PR #104, opened 2017-01-22
Debugging messages to debug release

owlishDeveloper:
A.S.:  All the review comments from the previous PR are included here.

> Trying to find the reason for intermittent failures of release event handling.
> 
> Handlers seem to be nicely equipped with debugging messages, so I focused on server side:
> 
> intree.js - check to see if YML is being parsed as expected; catch the taskGroupID for releases.
> api.js - check to see if release webhooks arrive and being recognized as such.
> exchanges.js - check to see if webhook payload is being parsed.
> 
> Also added a bit of clarity so that we instantly could see where the message comes from.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -1,4 +1,4 @@
-let debug = require('debug')('taskcluster-github');
+let debug = require('debug')('taskcluster-github:api');
 let crypto = require('crypto');
 let API = require('taskcluster-lib-api');
 let _ = require('lodash');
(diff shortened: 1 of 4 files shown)


djmitche (approved):
This looks great -- let's sanity-check on staging and then push to production!

djmitche:
Should we get this landed?

owlishDeveloper:
Let me check in staging one more time

owlishDeveloper:
@djmitche Yeah, we can now have this landed, ty!

PR #108, opened 2017-01-26
Working integration

owlishDeveloper:
Some debug messages and try-catches will go. The [integration](https://github.com/integration/tc-cat) is now being tested. Need a 60% review, please!


package.json
```diff
@@ -23,6 +23,7 @@
     "github": "^5.2.0",
     "js-yaml": "^3.6.1",
     "json-parameterization": "^0.2.0",
+    "jsonwebtoken": "https://registry.npmjs.org/jsonwebtoken/-/jsonwebtoken-5.7.0.tgz",
```
imbstack: Is there a reason we can't use the normal link for this?

schemas/constants.yml
```diff
@@ -12,6 +12,10 @@ identifier-pattern:     "^([a-zA-Z0-9-_]*)$"
 identifier-min-length:  1
 identifier-max-length:  22
 
+github-installation-pattern:  "^[0-9]+$"
+github-installation-min-length: 1
+github-installation-max-length: 10
```
imbstack: Maybe we can link to where this is defined in their docs or something.

schemas/github-pull-request-message.yml
```diff
@@ -19,6 +19,13 @@ properties:
     minLength:    {$const: github-identifier-min-length}
     maxLength:    {$const: github-identifier-max-length}
     pattern:      {$const: github-identifier-pattern}
+  installationID:
+    description: |
+       The installation which had an event.
+    type:          number
+    minLength:    {$const: github-installation-min-length}
+    maxLength:    {$const: github-installation-max-length}
```
imbstack: If this is a `number` we want `minumum` and `maximum` not `minLength` and `maxLength`. http://epoberezkin.github.io/ajv/keywords.html#keywords-for-numbers

schemas/github-pull-request-message.yml
```diff
@@ -19,6 +19,13 @@ properties:
...
+    pattern:      {$const: github-installation-pattern}
```
imbstack: I don't think we need a `pattern` if this is a number.

src/api.js
```diff
@@ -163,16 +165,19 @@ api.declare({
       msg.organization = sanitizeGitHubField(body.repository.owner.login),
       msg.action = body.action;
       msg.details = getPullRequestDetails(body);
+      msg.installationID = body.installation.id;
```
imbstack: These messages end up in a publicly readable stream on pulse. Are they something we should be keeping more secret? I'm thinking that we should probably do the following (although probably not in this PR):

1. Remove the installationId from this message and just publish it as it was before.
2. Don't have the worker in a separate process that is listening on pulse. Let's just do that work in the web endpoint. We can still publish messages in case anybody wants them, but it is not needed for the worker afaict.

src/exchanges.js
```diff
@@ -31,6 +31,11 @@ let commonRoutingKey = function(options) {
                         'characters aside from - and _ have been stripped.',
       maxSize:          100,
       required:         true,
+    }, {
+      name:             'installationID',
+      summary:          'The installation which had an event.',
```
imbstack: Maybe explain what an intstallation is here a bit more? Even saying something like "Github installation" might help.

src/main.js
```diff
@@ -15,6 +15,8 @@ let validator = require('taskcluster-lib-validate');
 let loader = require('taskcluster-lib-loader');
 let docs = require('taskcluster-lib-docs');
 let App = require('taskcluster-lib-app');
+let jwt = require('jsonwebtoken');
+let fs = require('fs');
```
imbstack: We don't use `fs` anymore, right?

imbstack (changes requested):
This is looking really good. I'm excited. Just tried out the integration [here](https://github.com/taskcluster/taskcluster-lib-monitor/pull/54) and it appeared to work perfectly!

package.json
owlishDeveloper: Oh, I don't know - this line happened after `npm install --save jsonwebtoken` . You mean, we should just replace this with a version number, as in other dependencies?

schemas/constants.yml
owlishDeveloper: It's not defined in the docs! I'll look again, but if it's not there, what do I do?

src/api.js
owlishDeveloper: OK, I'll do that! Thank you!

djmitche: Is that necessary?  It sounds like a lot of work, to not use pulse to separate webhooks from handlers.

I don't think the installation ID's are necessarily secret.  We *do* need to be careful that they cannot be forged, though.  I'll check that while reviewing.

src/api.js
```diff
@@ -182,12 +187,22 @@ api.declare({
...
+  }
+
+  // Not all webhook payloads include an e-mail for the user who triggered an event
+  let headUser = msg.details['event.head.user.id'].toString();
+  let userDetails = await instGithub.users.getById({id: headUser});
+  msg.details['event.head.user.email'] = userDetails.email || 
+    msg.details['event.head.user.login'] + '@users.noreply.github.com';
 
-  msg.details['event.head.user.email'] = userDetails.email || headUser + '@users.noreply.github.com';
+  msg.details['event.head.user.email'] = msg.details['event.head.user.login'] + '@users.noreply.github.com';
```
djmitche: Maybe my eyes are crossing, but it looks like you're seeting the same value twice here..

src/data.js
```diff
@@ -19,5 +19,6 @@ module.exports.Build = Entity.configure({
     state: Entity.types.String,
     created: Entity.types.Date,
     updated: Entity.types.Date,
+    installationID: Entity.types.Number,
```
djmitche: If adding a new column to an Azure table, you'll need to [make a new "version" of the table](https://github.com/taskcluster/azure-entities#migrations), rather than just changing it in place.

djmitche: Also, minor thing, but we tend to spell it "Id" rather than "ID" (e.g., taskId)

src/handlers.js
```diff
@@ -137,14 +147,30 @@ async function statusHandler(message) {
...
+    });
+  } catch (e) {
+    debug('Failed to modify the build!');
+    throw e;
+  }
+
+  // Authenticating as installation.
+  try {
+    debug('Authenticating as installation in status handler...');
+    var instGithub = await this.context.github.getInstallationGithub(build.installationID);
```
djmitche: This is good -- the installationId can't be forged here.

src/handlers.js
```diff
@@ -166,14 +192,26 @@ async function statusHandler(message) {
 async function jobHandler(message) {
   let context = this.context;
 
+  // Authenticating as installation.
+  try {
+    debug(`Authenticating as installation ${message.payload.installationID} 
+      to handle ${message.payload.details['event.type']}`);
```
djmitche: Here the installationId comes from the payload -- not from the routing key.  I think that's OK (we're assuming that only tc-github can publish to the pulse exchange, so nothing else can forge the message).  However, it does mean that we don't need the installationId in the routingKey.

src/handlers.js
```diff
@@ -289,18 +335,25 @@ async function jobHandler(message) {
...
+    await instGithub.repos.createCommitComment({
       owner: organization,
       repo: repository,
       sha,
-      body: 'Submitting the task to Taskcluster failed. ' +
-      '\nDetails:\n\n```js\n' +  errorBody + '\n```',
+      body: [
+        '<summary>Submitting the task to TaskCluster failed.',
+        '<details>\n',
+        '```js',
```
djmitche: Awesome using the js formatting here -- these errors are much more readable!!

src/intree.js
```diff
@@ -49,6 +49,7 @@ function completeInTreeConfig(config, payload) {
           GITHUB_HEAD_BRANCH: payload.details['event.head.repo.branch'],
           GITHUB_HEAD_REF: payload.details['event.ref'],
           GITHUB_HEAD_USER_EMAIL: payload.details['event.head.user.email'],
+          GITHUB_INSTALLATION_ID: payload.installationID,
```
djmitche: Do you know of a case where this will be useful?  Installations are just parts of an integration, and the created task is not related to the integration.  In other words, without the PEM, I think this value is not useful.  So there's probably no point in sending it to the task.

src/exchanges.js
djmitche: As I note below, I don't think this is necessary -- we can leave the routing keys as they are.

djmitche (changes requested):
Phew, that's a lot of notes for now.  Next round, I'll have some suggestions on cleaning up the code, but I've left that out for now since the debugging stuff is still useful.

src/api.js
djmitche: I don't see a forgery risk here.

imbstack: I want to do the work at some point to reduce the number of moving parts in tc-gh. There's no need for the indirection of having pulse in the middle and I'm 99% sure we got a random slowdown last week from having it there. That said, let's not do this work now!

owlishDeveloper:
I think I haven't miss anything! Thank you guys for the review!

package.json
djmitche: Yes, "^5.7.0"

schemas/constants.yml
djmitche: I think we can just count this as an integer, and not try to apply min/max length. So no need for these constants.

src/data.js
```diff
@@ -8,7 +8,7 @@ module.exports = {};
  *
  */
 module.exports.Build = Entity.configure({
-  version: 1,
+  version: 2,
```
djmitche: Upgrading the version requires that both versions 1 and 2 be in the source code -- the whole `.configure({..})` block needs to be copied.  So

```js
module.exports.Build = Entity.configure({
  version: 1,
  ...
}).configure({
  version: 2,
  ...
  installationId: Entity.types.Number,
  ...
});
```

src/handlers.js
```diff
@@ -116,16 +119,23 @@ module.exports = Handlers;
...
 async function statusHandler(message) {
   let taskGroupId = message.payload.taskGroupId || message.payload.status.taskGroupId;
-  debug(`handling state change for task-group ${taskGroupId}`);
+  debug(`Handling state change for task-group ${taskGroupId}`);
 
-  let build = await this.context.Builds.load({
-    taskGroupId,
-  });
+  try {
+    debug('Trying to load the build...');
```
djmitche: Has this section failed frequently?  If not, let's remove the try/catch and debugging.  An exception here will be handled by `callHandler`.

src/handlers.js
```diff
@@ -137,14 +147,30 @@ async function statusHandler(message) {
     state = 'failure';
   }
 
-  await build.modify((b) => {
-    b.state = state;
-    b.updated = new Date();
-  });
+  try {
```
djmitche: Same here -- the try/catch ends up making the code hard to read, and I don't think it's especially useful for debugging right now.

src/handlers.js
```diff
@@ -137,14 +147,30 @@ async function statusHandler(message) {
...
+    });
+  } catch (e) {
+    debug('Failed to modify the build!');
+    throw e;
+  }
+
+  // Authenticating as installation.
+  try {
+    debug('Authenticating as installation in status handler...');
+    var instGithub = await this.context.github.getInstallationGithub(build.installationId);
```
djmitche: Similarly here.  I think you could leave a `debug()` in here, but add the `installationId` to the message.

djmitche: Actually, it's probably best for `getInstallationGithub` to do that debug logging.  Then you can just call it here.

src/handlers.js
```diff
@@ -166,14 +192,26 @@ async function statusHandler(message) {
 async function jobHandler(message) {
   let context = this.context;
 
+  // Authenticating as installation.
+  try {
```
djmitche: Same here -- remove the try/catch and debug logging, letting `getInstallationGithub` do the logging.

src/handlers.js
```diff
@@ -236,17 +277,21 @@ async function jobHandler(message) {
...
       });
+      debug('Creating tasks...');
       await Promise.all(graphConfig.tasks.map(t => queue.createTask(t.taskId, t.task)));
 
       // We used to comment on every commit, but setting the status
       // is a nicer thing to do instead. It contains all of the same
       // information.
-      await context.github.repos.createStatus({
+
+      debug('Trying to create status...');
```
djmitche: Since we had some 404 errors with this call, can you add owner, repo, and sha to this log message?

src/handlers.js
djmitche: I switched this around in a commit to master -- The details element should surround the summary element.

src/main.js
```diff
@@ -74,14 +75,45 @@ let load = loader({
...
         promise: Promise,
       });
-      if (cfg.github.credentials.token) {
-        github.authenticate(cfg.github.credentials);
-      }
-      return github;
+
+      // This object insures that the authentication is delayed until we need it.
+      // Also, the authentication happens not just once in the beginning, but for each request.
+      return {
```
djmitche: This is pretty long to include in `main.js`.  Please move this code to a new module (`src/github_auth.js` or something like that).

src/main.js
```diff
@@ -74,14 +75,45 @@ let load = loader({
...
+            github.authenticate({type: 'integration', token: inteToken});
+            debug('Authenticated as integration');
+          } catch (e) {
+            debug('Authentication as integration failed!');
+            throw e;
+          }
+          var instaToken = await github.integrations.createInstallationToken({
+            installation_id: inst_id,
+          });
+          debug(`Obtained installation token: ${instaToken.token}`);
```
djmitche: We probably shouldn't log the token!  Logging secret values is generally a bad idea, as it's easy to copy/paste logs and accidentally disclose those secrets.  Instead, log the installationId so we know which installation was involved.

src/main.js
```diff
@@ -74,14 +75,45 @@ let load = loader({
...
+            {
+              iss: cfg.github.credentials.integrationId,
+            },
+            cfg.github.credentials.privatePEM,
+            {
+              algorithm: 'RS256',
+              expiresIn: '1m',
+            }
+          );
+          debug('Authenticating as integration...');
```
djmitche: This process will happen a lot, so let's limit it to one debug message per call to `getInstallationGithub` -- probably just the last one (for success)

djmitche (changes requested):
I made a bunch of comments on the production code.  I'll come back to look at tests after.

djmitche:
Oh, please also do a rebase due to the conflicts.  I think they'll be minor.

src/github_auth.js
```diff
@@ -0,0 +1,45 @@
...
+    getInstallationGithub: async (inst_id) => {
+      // Authenticating as integration
+      let inteToken = jwt.sign(
+        {iss: cfg.github.credentials.integrationId},
+        cfg.github.credentials.privatePEM,
+        {algorithm: 'RS256', expiresIn: '1m'},
+      );
+      try {
+        github.authenticate({type: 'integration', token: inteToken});
+        debug('Authenticated as integration');
```
djmitche: let's delete this debug, since it will be logged on every successful authentication and doesn't add much

src/github_auth.js
```diff
@@ -0,0 +1,45 @@
...
+
+      // Authenticating as installation
+      var instaToken = await github.integrations.createInstallationToken({
+        installation_id: inst_id,
+      });
+      debug(`Obtained installation token for installation: ${inst_id}`);
+      let gh = new Github({promise: Promise});
+      try {
+        gh.authenticate({type: 'token', token: instaToken.token});
+        debug('Authenticated as installation');
```
djmitche: ..and let's add the installationId here and remove the one above.  I'm just trying to shrink the success-case logging down to one line, since this will happen often.

src/data.js
```diff
@@ -30,10 +30,12 @@ module.exports.Build = Entity.configure({
...
     created: Entity.types.Date,
     updated: Entity.types.Date,
-    installationId: Entity.types.Number,
+    // GitHub installation ID that comes from the webhook
+    // Needed for authentication in statusHandler
+    installationId: Entity.types.Number, 
   },
   migrate: function(item) {
-    item.installationId = 1;
+    item.installationId = 0;
```
djmitche: +1

src/main.js
```diff
@@ -75,53 +73,7 @@ let load = loader({
...
-          } catch (e) {
-            debug('Authentication as integration failed!');
-            throw e;
-          }
-          
-          return gh;
-        },
-      };
-    },
+    setup: ({cfg}) => githubAuth(cfg),
```
djmitche: Looks good!

djmitche (approved):

imbstack (approved):
SO COOL

PR #115, opened 2017-02-06
Make different builds for different events.

imbstack:
This should fix the race conditions we see.

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -32,10 +32,27 @@ module.exports.Build = Entity.configure({
     updated: Entity.types.Date,
     // GitHub installation ID that comes from the webhook
     // Needed for authentication in statusHandler
-    installationId: Entity.types.Number, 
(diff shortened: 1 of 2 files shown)


src/data.js
```diff
@@ -32,10 +32,29 @@ module.exports.Build = Entity.configure({
...
+  properties: {
+    organization: Entity.types.String,
+    repository: Entity.types.String,
+    sha: Entity.types.String,
+    taskGroupId: Entity.types.String,
+    state: Entity.types.String,
+    created: Entity.types.Date,
+    updated: Entity.types.Date,
+    // GitHub installation ID that comes from the webhook
+    // Needed for authentication in statusHandler
```
djmitche: can drop these comments in version 3 since they are in version 2

djmitche (approved):

PR #120, opened 2017-02-06
Fix tests

djmitche:
Well, ok, mostly skipping them..


imbstack:
This seems to be consistently failing on travis, but otherwise looks good.

PR #138, opened 2017-02-09
Bug 1257540: allow customized pull requests policy

djmitche:


owlishDeveloper (approved):

src/pr-allowed.js
```diff
@@ -0,0 +1,109 @@
...
+    // No error, the user is a collaborator
+    debug(`Checking collaborator: ${login} is a collaborator on ${organization}/${repository}: True!`);
+    return true;
+  } catch (e) {
+    if (e.code == 404) {
+      // Only a 404 error means the user isn't a collaborator
+      // anything else should just throw like normal
+    } else if (e.code == 403) {
+      let msg = `Taskcluster does not have permission to check for repository collaborators.
+        Ensure that it is a member of a team with __write__ access to this repository!`;
```
imbstack: I guess this line isn't strictly true anymore.

imbstack: Really, I think this whole `else if` doesn't really make sense anymore in a post-taskclusterrobot world. We can probably get rid of it.

imbstack (approved):
Just a little nit and then 🚢

src/pr-allowed.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+    throw e;
+  }
+
+  // consult its `allowPullRequests` field
+  return taskclusterYml['allowPullRequests'] || DEFAULT_POLICY;
+}
+
+async function isCollaborator({login, organization, repository, sha, instGithub, debug}) {
+  if (login === organization) {
```
imbstack: I don't think this is necessary anymore. The owner should be a collaborator according to their docs. (and also now according to me trying it locally)

imbstack (approved):
Aha! r+ after nit again!

PR #144, opened 2017-02-14
New API endpoint to find out if the integration is set for the given repo

owlishDeveloper:
NEW DESCRIPTION:
1. Added another Azure table for storing integration owners and installation IDs, so we could look up an ID for a given owner.
2. Added integration_installation event handler to populate the above table.
3. Added the API endpoint which gets owner name and repo name, looks up installation ID in the above table, and uses it to check if the integration was installed for the given repo (using github API). 
4. Added output schema for the API
5. Added three tests for the thing.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -198,6 +198,12 @@ api.declare({
       msg.details = getReleaseDetails(body);
       msg.installationId = body.installation.id;
       publisherKey = 'release';
+    } else if (eventType == 'integration_installation') {
+      await this.context.OwnersDirectory.create({
+        installationId: body.installation.id,
+        owner: body.installation.account.login,
+      });
+      return resolve(res, 200, 'Created table row!');
     } else {
       return resolve(res, 400, 'No publisher available for X-GitHub-Event: ' + eventType);
     }
@@ -271,4 +277,37 @@ api.declare({
       };
     }),
   });
+});
+
+api.declare({
+  name: 'isInstalledFor',
+  title: 'Check if Repository has Integration',
+  description: [
+    'Checks if the integration has been installed for',
+    'a given repository of a given organization or user.',
+  ].join('\n'),
+  stability: 'experimental',
+  method: 'get',
+  route: '/repository/:owner/:repo',
+  output: 'is-installed-for.json',
+}, async function(req, res) {
+  // Extract owner and repo from request into variables
+  let {owner, repo} = req.params;
+
+  // Look up the installation ID in Azure. If no such owner in the table, no error thrown
+  let ownerInfo = await this.OwnersDirectory.load({owner}, true);
+
+  if (ownerInfo) {
+    let instGithub = await this.github.getInstallationGithub(ownerInfo.installationId);
+    let reposList = await instGithub.integrations.getInstallationRepositories();
+
+    // GitHub API returns an array of objects, each of wich has an array of repos
+    let installed = reposList.reduce((a, b) => a.repositories.concat(b.repositories), {repositories: []})
+      .map(repo => repo.name)
+      .indexOf(repo);
+
+    res.reply({installed: installed != -1});
+  }
+    
+  res.reply({installed: false});
 });
--- a/src/data.js
+++ b/src/data.js
@@ -74,3 +74,13 @@ module.exports.Build = Entity.configure({
     return item;
   },
 });
+
+module.exports.OwnersDirectory = Entity.configure({
+  version: 1,
+  partitionKey: Entity.keys.StringKey('owner'),
+  rowKey: Entity.keys.ConstantKey('someConstant'),
+  properties: {
+    installationId: Entity.types.Number,
+    owner: Entity.types.String,
+  },
+});
--- a/test/api_test.js
+++ b/test/api_test.js
@@ -8,9 +8,15 @@ suite('api', () => {
   let assert = require('assert');
   let _ = require('lodash');
 
+  let github = Object.create(null);
+
   suiteSetup(async () => {
     await helper.Builds.scan({}, {
       handler: build => build.remove(),
+    });
+
+    await helper.OwnersDirectory.scan({}, {
+      handler: owner => owner.remove(),
     });
 
     await helper.Builds.create({
@@ -61,6 +67,21 @@ suite('api', () => {
       eventType: 'push',
       eventId: '26370a80-ed65-11e6-8f4c-80082678482d',
     });
+
+    await helper.OwnersDirectory.create({
+      installationId: 9090,
+      owner: 'abc123',
+    });
+
+    await helper.OwnersDirectory.create({
+      installationId: 9091,
+      owner: 'qwerty',
+    });
+  });
+
+  setup(async () => {
+    github = await helper.load('github');
+    github.inst(9090).setRepositories('coolRepo', 'anotherCoolRepo', 'awesomeRepo');
   });
 
   test('all builds', async function() {
@@ -99,5 +120,14 @@ suite('api', () => {
     assert.equal(builds.builds[0].organization, 'abc123');
     assert.equal(builds.builds[0].repository, 'xyz');
     assert.equal(builds.builds[0].sha, 'y650871208002a13ba35cf232c0e30d2c3d64783');
+  });
+
+  test('integration installation', async function() {
+    let result = await helper.github.isInstalledFor('abc123', 'coolRepo');
+    assert.deepEqual(result, {installed: true});
+    result = await helper.github.isInstalledFor('abc123', 'unknownRepo');
+    assert.deepEqual(result, {installed: false});
+    result = await helper.github.isInstalledFor('unknownOwner', 'unknownRepo');
+    assert.deepEqual(result, {installed: false});
   });
 });
--- a/test/github-auth.js
+++ b/test/github-auth.js
@@ -10,6 +10,7 @@ class FakeGithub {
     this.repo_collaborators = {};
     this.github_users = [];
     this.repo_info = {};
+    this.repositories = [];
 
     const throwError = code => {
       let err = new Error();
@@ -64,6 +65,9 @@ class FakeGithub {
           throwError(404);
         }
       },
+      'integrations.getInstallationRepositories': async() => {
+        return this.repositories;
+      },
     };
 
     const debug = Debug('FakeGithub');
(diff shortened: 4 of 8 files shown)


src/data.js
```diff
@@ -74,3 +74,13 @@ module.exports.Build = Entity.configure({
     return item;
   },
 });
+
+module.exports.OwnersDirectory = Entity.configure({
+  version: 1,
+  partitionKey: Entity.keys.StringKey('owner'),
+  rowKey: Entity.keys.ConstantKey('someConstant'),
+  properties: {
+    installationID: Entity.types.Number,
```
djmitche: Typically we would spell this `installationId` (I know, really minor detail...)

If this is easy to change now, let's fix it.  If it reqiures a new `.configure({version: 2, ..})` then it's not worth ith.

test/api_test.js
```diff
@@ -9,10 +9,16 @@ suite('api', () => {
   let _ = require('lodash');
 
   suiteSetup(async () => {
+    helper.github.inst(9091).setRepositories('coolRepo', 'anotherCoolRepo', 'awesomeRepo');
```
djmitche: awesomeRepo is my favorite, the code in it is totally sweet

test/api_test.js
```diff
@@ -100,4 +116,9 @@ suite('api', () => {
     assert.equal(builds.builds[0].repository, 'xyz');
     assert.equal(builds.builds[0].sha, 'y650871208002a13ba35cf232c0e30d2c3d64783');
   });
+
+  test('integration installation', async function() {
+    let result = await helper.github.isRepoSet('qwerty', 'coolRepo');
+    assert.deepEqual(result, {installed: true});
+  });
```
djmitche: This suite should probably test the other two cases, too:
 * no such org known
 * org is known but the repo is not installed

test/github-auth.js
```diff
@@ -100,6 +105,11 @@ class FakeGithub {
   setUser({id, email}) {
     this.github_users.push({id: id.toString(), email});
   }
+
+  setRepositories(repoName) {
+    let repos = [...repoName].map(repo => {name: repo});
```
djmitche: I think `repoName` is actually an array of names, right?
```
> repoName = "awesomeRepo";
'awesomeRepo'
> [...repoName]
[ 'a', 'w', 'e', 's', 'o', 'm', 'e', 'R', 'e', 'p', 'o' ]
```
maybe just change the name of the variable so the reader doesn't get confused?

test/github-auth.js
```diff
@@ -55,6 +57,9 @@ class FakeGithub {
           throwError(404);
         }
       },
+      'integrations.getInstallationRepositories': async() => {
+        return this.repositories[installation_id];
+      },
```
djmitche: I saw your ping about this last night, but it looks like you figured it out.  This works great.

Two minor cleanups though:
 * `this.installation_id` is never used (instead, you use `installation_id` directly here, which is fine)
 * We make a new `FakeGithub` instance for each installation, so `this.repositories` will only ever have one property, the installation ID.  So you could simplify it a little by not indexing it with the installation ID:
```
      'integrations.getInstallationRepositories': async() => {
        return this.repositories;
      },
```
and
```

  setRepositories(repoName) {
    this.repositories = {repositories: [...repoName].map(repo => {name: repo})}; 
  }
```

src/api.js
```diff
@@ -272,3 +278,36 @@ api.declare({
     }),
   });
 });
+
+api.declare({
+  name: 'isRepoSet',
```
djmitche: Would `isInstalledFor` be a better name?

djmitche (approved):
This looks great, reads clearly, has good tests, and from everything I can see works perfectly.  I've made a few comments for minor improvements that I'd like to see addressed, but marked it as "approved" all the same since I'd also be fine with this landing without any further changes.

src/data.js
owlishDeveloper: Oi wei 😱 You've already told me about that id spelling, and I forgot, it's not good.
(Does it really work perfectly for you though? I rebased the branch, but still getting errors...)

test/api_test.js
owlishDeveloper: Yeah, I planned those... I have these test cases: "Creating an entity correctly", "Returning true for existing repo", "Returning false for absent repo", "Returning false for absent owner", "Notifying user about invalid input". Are those good? Anything else to add?

test/github-auth.js
owlishDeveloper: Cool! Thanks!

test/github-auth.js
owlishDeveloper: oi wei 😱 I forgot another spread!

test/github-auth.js
```diff
@@ -114,6 +118,13 @@ class FakeGithub {
   setUser({id, email}) {
     this.github_users.push({id: id.toString(), email});
   }
+
+  setRepositories(...repoName) {
+    this.repositories.push({
+      repositories: [...repoName].map(repo => {return {name: repo};}),
+    });
+    console.log('REPOS:', this.repositories, this.repositories[0].repositories[0].name);
```
djmitche: If you use `console.log(JSON.stringify(this.repositories))` you can see the whole data structure -- `console.log` gives up after a while and just prints `[Object]`.  Here, I get
```
REPOS: [{"repositories":[{"name":"coolRepo"},{"name":"anotherCoolRepo"},{"name":"awesomeRepo"}]}]
```
which looks right to me.

src/api.js
```diff
@@ -272,3 +278,44 @@ api.declare({
...
+  // Look up the installation ID in Azure. If no such owner in the table, no error thrown
+  let ownerInfo = await this.OwnersDirectory.load({owner}, true);
+
+  if (ownerInfo) {
+    let instGithub = await this.github.getInstallationGithub(ownerInfo.installationId);
+    let reposList = await instGithub.integrations.getInstallationRepositories();
+
+    // GitHub API returns an array of objects, each of wich has an array of repos
+    // [ { repositories: [ [Object], [Object], [Object] ] } ]
+    console.log('reposList', reposList);
```
djmitche: So the question is, why is this printing `[]`?

Ah, this one took me a bit!  The problem is that you're calling `github.inst(9090).setRepositories('coolRepo', 'anotherCoolRepo', 'awesomeRepo');` in `suiteSetup`, which only runs once before any of the tests in that suite.  But the helper resets the stubs (deleting this nice setup data) before every test:

```js
mocha.beforeEach(async () => {
  let github = await helper.load('github');
  github.resetStubs();
```

if you move the `setRepositories` call to a setup method in api_test.js:

```js
setup(async () => {
  github = await helper.load('github');
  github.inst(9090).setRepositories('coolRepo', 'anotherCoolRepo', 'awesomeRepo');
});
```

it should get the right value.

src/api.js
```diff
@@ -272,3 +278,44 @@ api.declare({
...
+    try {
+      let installed = reposList.reduce((a, b) => a.repositories.concat(b.repositories), {repositories: []})
```
djmitche: When `reposList` is `[]`, this does something different than when it's actually a list of repositories.  That worries me, even though it's unlikely to happen in real life.  That's why you're getting "reposList.reduce(...).map is not a function", I think.

What about this:

```js
reposList.reduce((a, b) => ({repositories: a.repositories.concat(b.repositories)}), {repositories: []})
  .repositories
  .map(repo => repo.name)
  .includes(repo);
```

This just makes the reducer function return `{repositories: [..]}` instead of returning `[..]` directly.  Done this way, its inputs and output all have the same shape (`{repositories: [..]}`), so whether we have an empty `reposList`, and thus return the initial value, or do a bunch of reductions, the shape always stays the same.  In general, this is a good principle when using reduce.

src/api.js
```diff
@@ -272,3 +278,44 @@ api.declare({
...
+      .map(repo => repo.name)
+      .includes(repo);
```
djmitche: And the last bit I had trouble with.. I boiled this down to

```js
assert(['a', 'b'].includes('a'));
```

which fails:
```
TypeError: ["a","b"].includes is not a function
```

John once told me something I didn't totally understand about babel-compile not supporting `.includes` on arrays, so I think that's what's going on here.  Is there another approach you could take to this test?

test/github-auth.js
owlishDeveloper: Ah, yes, of course! Thanks!

src/api.js
owlishDeveloper: I did `console.log(['a', 'b'].includes('a'));` in api.js, and it printed out true. So, it seems like Babel supports `includes`. However, in ye olde ES5 this problem, I think, would typically be solved with Array.prototype.indexOf(), which returns -1 if the element is absent. So, it can be:

```
let installed = reposList
  .reduce((a, b) => a.repositories.concat(b.repositories), {repositories: []})
  .map(repo => repo.name)
  .indexOf(repo)
res.reply({installed: installed != -1})
```
for example

src/api.js
owlishDeveloper: Hmm. That's what I started my morning with, I ran in Node:
```
let d = []
d.reduce((a, b) => a.repositories.concat(b.repositories), {repositories: []})
```
It produced `{ repositories: [] }`

However, if the structure of the incoming array is different (like, [1, 2] instead of [{repositories: []}] - then it would produce an array with undefined neatly put inside

I'll think about this, though! It's interesting, thank you!

owlishDeveloper:
This can be now merged, I think. @djmitche and @imbstack - can you please take a final look?

src/api.js
```diff
@@ -198,6 +198,12 @@ api.declare({
       msg.details = getReleaseDetails(body);
       msg.installationId = body.installation.id;
       publisherKey = 'release';
+    } else if (eventType == 'integration_installation') {
+      await this.context.OwnersDirectory.create({
+        installationId: body.installation.id,
+        owner: body.instalation.account.login,
```
djmitche: I just spotted this -- spelling error in "installation"

djmitche (approved):
Looks good with that one spelling fix.

We had talked earlier about filling this table for existing integrations by populating it on other requests, too.  Maybe that can be a follow-up PR?

One more thing: what happens if we get two installation calls for the same owner?  We could address that in a followup too..

src/api.js
owlishDeveloper: 😱

owlishDeveloper:
Travis is getting really weird here

djmitche:
Let me know when you're around Monday, and we can test this in staging.

owlishDeveloper:
Ok!

PR #146, opened 2017-02-20
Badges API v1.0

owlishDeveloper:


src/api.js
```diff
@@ -280,6 +280,53 @@ api.declare({
...
+    try {
+      let instGithub = await this.github.getInstallationGithub(ownerInfo.installationId);
+      // Statuses is an array of status objects
+      let statuses = await instGithub.repos.getStatuses({owner, repo, ref: branch});
+      let status = statuses.find(statusObject => statusObject.context.startsWith('Taskcluster'));
+
+      // The weird switch is here for the future changes
+      switch (status.state) {
+        case 'success':
+          res.reply({status: 'success'});
```
djmitche: This looks like a JSON object, not an image :)

src/api.js
```diff
@@ -280,6 +280,53 @@ api.declare({
...
+
+  // Look up the installation ID in Azure. If no such owner in the table, no error thrown
+  let ownerInfo = await this.OwnersDirectory.load({owner}, true);
+
+  if (ownerInfo) {
+    try {
+      let instGithub = await this.github.getInstallationGithub(ownerInfo.installationId);
+      // Statuses is an array of status objects
+      let statuses = await instGithub.repos.getStatuses({owner, repo, ref: branch});
+      let status = statuses.find(statusObject => statusObject.context.startsWith('Taskcluster'));
```
djmitche: Rather than hardcode "Taskcluster" here, this should probably use the same configuration as we use to generate the status context.

src/api.js
```diff
@@ -280,6 +280,53 @@ api.declare({
...
+          res.reply({status: 'success'});
+          break;
+        case 'failure':
+          res.reply({status: 'failure'});
+          break;
+        case 'pending':
+          res.reply({status: 'pending'});
+          break;
+        default:
+          res.reply({status: 'tc spirit is strong with this one'});
```
djmitche: Let's just return 404 for this, too..

djmitche (changes requested):
As for the commits from the other branch, a rebase onto master should fix that up.

src/api.js
djmitche: http://stackoverflow.com/questions/17515699/node-express-sending-image-files-as-api-response

You'll also need to set the content-type:
```
     res.writeHead(200, {'Content-Type': 'image/png' });
```

src/api.js
owlishDeveloper: done

src/api.js
owlishDeveloper: (where will the images be coming from?)

src/api.js
owlishDeveloper: done

src/api.js
djmitche: I think you'll need to create some, perhaps based on emoji.  You could generate them in inkscape (svg) or gimp (png).  Then just include the image files in the repo, and serve them using `res.sendFile`.

owlishDeveloper: Should they be in a separate folder or in the root?

djmitche: How about in `assets/`.

owlishDeveloper: oh darn, just saw your reply. ok, I'll make assets/ (in the root  this one, right?)

djmitche: yes

src/api.js
```diff
@@ -280,6 +280,55 @@ api.declare({
...
+      // The weird switch is here for the future changes
+      switch (status.state) {
+        case 'success':
+          res.writeHead(200, {'Content-Type': 'image/png'});
+          res.reply({status: 'success'});
+          break;
+        case 'failure':
+          res.writeHead(200, {'Content-Type': 'image/png'});
+          res.sendFile('../assets/failure.png');
+          break;
```
djmitche: This looks right to me.. let's see if it works :)

owlishDeveloper: I'm not sure how to do that :) But don't give me hints just yet, I'd like to try and think of something

owlishDeveloper:
The tests pass for me, but here they will fail because I have to add path to images to config (or maybe that's the wrong way, I'm not sure). Not sure what the path is exactly... https path to the images here in the repo?

And another thing - I don't know what to do if it gets request for owner/repo it doesn't know. More in the comments there in the code

owlishDeveloper:
(you probably will not like the success emoji, but I had to try)

owlishDeveloper:
Also, the badges will be edited, here are just raw emojis

src/api.js
```diff
@@ -280,6 +280,45 @@ api.declare({
...
+
+  // Look up the installation ID in Azure. If no such owner in the table, no error thrown
+  let ownerInfo = await this.OwnersDirectory.load({owner}, true);
+
+  if (ownerInfo) {
+    try {
+      let instGithub = await this.github.getInstallationGithub(ownerInfo.installationId);
+      // Statuses is an array of status objects, where we find the relevant status
+      let statuses = await instGithub.repos.getStatuses({owner, repo, ref: branch});
+      let status = statuses.find(statusObject => statusObject.context.startsWith(`${this.cfg.app.statusContext}`));
```
imbstack: Perhaps the proper way to do this is check that the user who created it is us? It has a user id associated with it, right?

imbstack (commented):
Otherwise this looks perfect!

src/api.js
owlishDeveloper: Yeah, there's 'creator' key, and inside there is 'taskcluster[bot]' as login and the id. Sure, I can do that

owlishDeveloper: (id can be exposed in the code, can't it? I can use login otherwise)

README.md
```diff
@@ -1,5 +1,6 @@
 TaskCluster GitHub Service
 ==========================
+https://github.taskcluster.net/v1/badge/taskcluster/taskcluster-github/badgers
```
djmitche:
```
![Build Status](https://github.taskcluster.net/v1/badge/taskcluster/taskcluster-github/badgers)
```
https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#images

src/api.js
```diff
@@ -281,6 +281,45 @@ api.declare({
 });
 
 api.declare({
+  name: 'badges',
```
djmitche: Probably singular would be better (it's only one badge..)

src/api.js
```diff
@@ -281,6 +281,45 @@ api.declare({
...
+
+  // Look up the installation ID in Azure. If no such owner in the table, no error thrown
+  let ownerInfo = await this.OwnersDirectory.load({owner}, true);
+
+  if (ownerInfo) {
+    try {
+      let instGithub = await this.github.getInstallationGithub(ownerInfo.installationId);
+      // Statuses is an array of status objects, where we find the relevant status
+      let statuses = await instGithub.repos.getStatuses({owner, repo, ref: branch});
+      let status = statuses.find(statusObject => statusObject.creator.id === 25589918);
```
djmitche: Is there a way we can learn this creator ID at runtime?  If not, can you explain how a user can figure it out, and put it in the configuration?

src/api.js
```diff
@@ -281,6 +281,45 @@ api.declare({
...
+
+      // Then we send a corresponding image. The path should be absolute
+      res.sendFile(this.cfg.server.publicUrl + '/assets/' + status.state + '.svg');
+    } catch (e) {
+      if (e.code < 500) {
+        // ???
+        return res.sendStatus(204);
```
djmitche: Maybe we need a special "error" badge that gets returned in this case?

src/api.js
```diff
@@ -281,6 +281,45 @@ api.declare({
...
+      }
+      throw e;
+    }
+  } else {
+    // ???
+    return res.sendStatus(204);
```
djmitche: The error badge would work here, too..

test/api_test.js
```diff
@@ -130,4 +143,25 @@ suite('api', () => {
     result = await helper.github.isInstalledFor('unknownOwner', 'unknownRepo');
     assert.deepEqual(result, {installed: false});
   });
+
+  test('build badges', async function() {
+    await request.get('http://localhost:60415/v1/badge/abc123/coolRepo/master').end((err, res) => {
+      err ? console.log(err) : assert.equal(res.headers['content-length'], 928);
```
djmitche: If there's an error, it will get sent to the console -- but the test will still succeed.  It's probably better to throw the error if it occurs.

Also, this is an odd combination of callback and `await`.  @imbstack I think you know more about superagent-promise than I do? Is this right?

test/api_test.js
```diff
@@ -130,4 +143,25 @@ suite('api', () => {
...
+
+    // This is one option for absent things:
+    await request.get('http://localhost:60415/v1/badge/abc123/unknownRepo/master').end((err, res) => {
+      err ? console.log(err) : assert.equal(res.statusCode, 204);
+    });
+    // Here I had another:
+    /*
+    The API returned 404 instead of 204, and here I had try-catch statement because 404 causes
+    superagent to throw an error. On superagent docs, there's request.get().ok(res => res.status < 500).end()
+    trick, but it didn't work here ('.ok is not a function').
```
djmitche: Maybe the difference between superagent and superagent-promise?  I'm not sure..

djmitche (commented):
Looks good so far!!

src/api.js
owlishDeveloper: Yeah, there is a way, I'm implementing it right now

src/api.js
owlishDeveloper: Yeah, that seems like a good idea. @imbstack, what are your thoughts?

test/api_test.js
owlishDeveloper: I tried connecting superagent itself - didn't work either :(

README.md
owlishDeveloper: Thank you!

src/api.js
owlishDeveloper: done

imbstack:
Honest question:

Should `assets/` be `src/assets/`? What's the proper place to put assets in an express project like this? I think it will be the first time we do it in taskcluster.

owlishDeveloper:
My understanding, there should be /public/img or smth to that effect in the root.

djmitche:
If it's in src/, it will be copied by babel-compile, which seems wasteful.  Based on my recent survey of all of our apps, I don't think it really matters :)

owlishDeveloper:
Ok, so I implemented everything you guys suggested; also, I added a real success badge to test. If no further changes requests, this can be merged and deployed somewhere.

src/api.js
```diff
@@ -281,6 +281,48 @@ api.declare({
...
+
+      let taskclusterBot = await instGithub.users.getForUser({username: 'taskcluster[bot]'});
+      // Statuses is an array of status objects, where we find the relevant status
+      let statuses = await instGithub.repos.getStatuses({owner, repo, ref: branch});
+      let status = statuses.find(statusObject => statusObject.creator.id === taskclusterBot.id);
+
+      // Then we send a corresponding image. (The path should be absolute)
+      return res.sendFile(status.state + '.svg', fileConfig);
+    } catch (e) {
+      if (e.code < 500) {
```
imbstack: I think in this case we also want to return this image if there's a 400.

The logic here is probably more like

```js
res.sendFile('error.svg', fileConfig);
if (e.code < 500) {
  throw e;
}
```

test/api_test.js
imbstack: Yeah, it's a little bit weird. We don't want to `console.log`. I'll comment here again soon when I figure out how to do this.

imbstack (commented):
Other than these little nits, I think the only thing we need to do to get this ready is to change `error.svg` to something like `unknown.svg` and make it neutral instead of bad. While we're at it, we may as well settle on which images we like for success/pending/failure.

test/api_test.js
owlishDeveloper: What if I put another assert.equal in there? For status code, for example. Then the test will fail in case of an error

src/api.js
owlishDeveloper: done

owlishDeveloper:
@imbstack I think this can be merged now!

README.md
```diff
@@ -8,6 +10,11 @@ This service monitors all of the repositories associated with an organization fo
 ### Docs
 See: https://docs.taskcluster.net/manual/vcs/github
 
+### Adding status badges to your project's readme:
+Insert the following string to the readme file in your project's repository:
+`![Task Status](https://github.taskcluster.net/v1/badge/taskcluster/taskcluster-github/badgers)`
```
djmitche: You might want to indicate where they should fill in the parts of their repo :)

README.md
```diff
@@ -40,3 +47,9 @@ run this check (but, no other credentials!)
 
 After you confirm it works on stage, deploy to production with [the
 pipeline](https://dashboard.heroku.com/pipelines/b867da9f-e443-4ddd-b8b1-2209532897b4).
+
+## Copyright notes
```
djmitche, line 51: :+1:

README.md
owlishDeveloper: done!

djmitche (approved):

Issue #148, opened 2017-02-21
YAMLException: duplicated mapping key at line 315, column 7:


imbstack:
https://sentry.io/taskcluster/taskcluster-github/issues/222528905/

```
YAMLException: duplicated mapping key at line 315, column 7:
          mounts:
          ^
  File "/app/lib/handlers.js", line 241, in Handlers._callee8$
    c = yaml.safeLoad(repoconf);
...
(17 additional frame(s) were not displayed)

YAMLException: duplicated mapping key at line 315, column 7:
          mounts:
          ^
```

petemoore:
Consistently occurs - here is my next push.

Thanks Brian!

https://papertrailapp.com/systems/taskcluster-github/events?focus=770778671832305677&selected=770778671832305677

https://github.com/taskcluster/generic-worker/blob/fb14121f8f16d8aad5da3bd71ebe0c0aebc7fc75/.taskcluster.yml#L315-L316

djmitche:
The error itself is invalid, as far as I can see -- the yaml looks fine.

I think the next debugging step is to pull that file and try running it through the version of the yaml lib we're using, and see if that raises a similar error.

petemoore:
credits to @imbstack  for finding the root cause - it was actually a combination of there being a duplicate elsewhere in the file ([here](https://github.com/taskcluster/generic-worker/blob/fb14121f8f16d8aad5da3bd71ebe0c0aebc7fc75/.taskcluster.yml#L284) and [here](https://github.com/taskcluster/generic-worker/blob/fb14121f8f16d8aad5da3bd71ebe0c0aebc7fc75/.taskcluster.yml#L311)) combined with this [reporting error](https://github.com/nodeca/js-yaml/issues/243).

Thanks guys, and sorry for not spotting the duplication myself!

djmitche:
This is still occurring, and I've verified that #149 is merged to production..

petemoore:
That's strange, I fixed the `.taskcluster.yml` file in generic-worker.

imbstack:
Ah, it's happening in pr-allowed now, instead of in handlers. The nice thing is that the library upgrade worked and is saying that the issue is duplicate artifacts now.

I expect the reason it is failing again in the case of pr-allowed is that the .taskcluster.yml is being loaded from the default branch (probably master?) and it has a duplicate mapping in it.

The fixes for this are twofold:

1. The default branch .taskcluster.yml needs fixed in generic-worker
[...]

PR #149, opened 2017-02-21
Report yaml error to users

imbstack:
This fixes one issue that arose in #148. This should probably be reported to the user rather than sentry. Next we have to figure out what's actually going wrong.


owlishDeveloper (approved):

imbstack:
Ok, this second diff adds a fix for #148 itself. It was due to nodeca/js-yaml#243. The yaml was invalid, but not in the way the library was telling us.

djmitche (approved):

PR #157, opened 2017-02-22
Handle EntityAlreadyExists error

owlishDeveloper:
If we get EntityAlreadyExists error, the entity is loaded and modified.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -199,10 +199,11 @@ api.declare({
       msg.installationId = body.installation.id;
       publisherKey = 'release';
     } else if (eventType == 'integration_installation') {
+      // Creates a new entity or overwrites an existing one
       await this.OwnersDirectory.create({
         installationId: body.installation.id,
         owner: body.installation.account.login,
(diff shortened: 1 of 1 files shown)


imbstack:
Is it possible we want to specify `true` as the second option to create instead of this try-catch? https://github.com/taskcluster/azure-entities#row-operations

owlishDeveloper:
Natürlich!

src/api.js
```diff
@@ -199,10 +199,13 @@ api.declare({
       msg.installationId = body.installation.id;
       publisherKey = 'release';
     } else if (eventType == 'integration_installation') {
+      // Creates a new entity or overwrites an existing one
       await this.OwnersDirectory.create({
         installationId: body.installation.id,
         owner: body.installation.account.login,
-      });
+      }, true);
+    }
```
imbstack: We don't want this extra `}` here, right?

imbstack (approved):

PR #160, opened 2017-02-27
Add support for users.getForUser GitHub API

owlishDeveloper:
I need this for badges. Badges make use of users.getForUser GitHub API. I think the best thing would be to add "username" key to user objects in this.github_users of FakeGitHub, and to re-use the existing setUser method.

### DIFF
--- a/test/github-auth.js
+++ b/test/github-auth.js
@@ -68,6 +68,15 @@ class FakeGithub {
       'integrations.getInstallationRepositories': async() => {
         return this.repositories;
       },
+      'users.getForUser': async ({username}) => {
+        let user = _.find(this.github_users, {username});
+        if (user) {
+          user.id = parseInt(user.id, 10);
+          return user;
+        } else {
+          throwError(404);
+        }
+      },
     };
 
     const debug = Debug('FakeGithub');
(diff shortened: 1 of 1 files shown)


test/github-auth.js
```diff
@@ -68,6 +68,15 @@ class FakeGithub {
       'integrations.getInstallationRepositories': async() => {
         return this.repositories;
       },
+      'users.getForUser': async ({username}) => {
+        let user = _.find(this.github_users, {username});
+        if (user) {
+          user.id = parseInt(user.id, 10);
```
djmitche, line 74: Why `parseInt` here after the explicit `toString` call below?

djmitche (approved):
I like nice, small PRs :)

test/github-auth.js
owlishDeveloper: Well, in real github payloads id is always a number. users.getForId uses id in form of a string - I have no idea why :) That's, I suspect, where toString comes from? ;) But for users.getForUser I need a number.

owlishDeveloper: I think I should add a comment about that in the code

djmitche: It looks like `user.getForId` returns the value as an integer (https://api.github.com/users/djmitche).  And `setUser` is just a test function, so if you always pass it an integer, there's no need to do anything.

djmitche: Oh, I see, users.getById (which already existed) searches by string.  OK, that makes sense then.

owlishDeveloper: Yeah, I meant get**By**Id, sorry for confusion :)

PR #164, opened 2017-03-02
Add a function to sync installationIds

imbstack:
That ended up being a bit more work than expected. However, the actual logic is quite slim. The other changes were to FakeGithub and github-auth in general. The new function in there allows us to grab the github client that we've configured without using an installation github, since that's what we need for this table. Otherwise, I switched our promise library to bluebird because it has that nice `Promise.map` thing.

To be honest, the existence of this function makes me wonder if we need a table for this stuff at all. Perhaps we should just hit this function directly every time we get a request to `isInstalled`? We can do that some other time though.


src/github-auth.js
```diff
@@ -1,37 +1,44 @@
...
-let Promise = require('promise');
+let Promise = require('bluebird');
 let jwt = require('jsonwebtoken');
 
 module.exports = async ({cfg}) => {
   let github = new Github({
     promise: Promise,
   });
 
+  let setupToken = async _ => {
```
owlishDeveloper: Just curious: what's this underscore syntax?

owlishDeveloper:
If there's no difference in performance, I agree that we don't need the table then.

owlishDeveloper:
(If we decide not to use the table, I would change the API as soon as pos. @djmitche what are your thoughts?)

owlishDeveloper (approved):

src/github-auth.js
```diff
@@ -1,37 +1,44 @@
...
-
+      await setupToken();
       // Authenticating as installation
       var instaToken = await github.integrations.createInstallationToken({
         installation_id: inst_id,
       });
       let gh = new Github({promise: Promise});
       try {
-        gh.authenticate({type: 'token', token: instaToken.token});
+        await gh.authenticate({type: 'token', token: instaToken.token});
```
djmitche: I don't think this is async -- it's just changing state on the Github object.

test/helper.js
```diff
@@ -82,6 +82,11 @@ mocha.before(async () => {
   helper.intree = overwrites.intree = await helper.load('intree', overwrites);
   webServer = overwrites.server = await helper.load('server', overwrites);
 
+  helper.install = (id, name, repos) => overwrites.github.createInstall(id, name, repos);
```
djmitche: I would rather provide the tests with a way to access the fake github object, and have them call `createInstall` on that object.  I think `helper.install` is a little too vague - install what?

src/main.js
```diff
@@ -144,6 +145,20 @@ let load = loader({
     }),
   },
 
+  syncInstallations: {
+    requires: ['github', 'OwnersDirectory'],
+    setup: async ({github, OwnersDirectory}) => {
+      let gh = await github.getIntegrationGithub();
+      let installations = await gh.integrations.getInstallations({});
```
djmitche, line 152: We chose not to call this function on every `isRepoInstalled` call because it may someday contain 100's of installations.  If github eventually adds a `getInstallation(owner, repo)` method, we could use that.

djmitche (changes requested):

PR #170, opened 2017-03-06
Changed the sizes of the images

owlishDeveloper:
This PR also contains new-sized images (the ones that align with the badges from Identical land).


owlishDeveloper:
weird

PR #175, opened 2017-03-07
Revert "Debugging message added"

djmitche:
Reverts taskcluster/taskcluster-github#173 (for when the debugging is done)


owlishDeveloper (approved):

owlishDeveloper:
This can be merged now, because that debugging message doesn't help. I'll think what else I can do about the problem.

PR #179, opened 2017-03-13
TypeError: Cannot read property 'state' of undefined

djmitche:
https://sentry.io/taskcluster/taskcluster-github/issues/228497262/

```
TypeError: Cannot read property 'state' of undefined
  File "/app/lib/api.js", line 475, in Object._callee3$
    return _context3.abrupt('return', res.sendFile(status.state + '.svg', fileConfig));
...
(1 additional frame(s) were not displayed)

TypeError: Cannot read property 'state' of undefined
```

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -314,8 +314,13 @@ api.declare({
       let statuses = await instGithub.repos.getStatuses({owner, repo, sha: branch});
       let status = statuses.find(statusObject => statusObject.creator.id === taskclusterBot.id);
 
-      // Then we send a corresponding image.
-      return res.sendFile(status.state + '.svg', fileConfig);
+      if (status) {
+        // If we got a status, send a corresponding image.
+        return res.sendFile(status.state + '.svg', fileConfig);
+      } else {
+        // otherwise, it's a commit without a TC status, which probably means a new repo
+        return res.sendFile('newrepo.svg', fileConfig);
+      }
     } catch (e) {
       if (e.code < 500) {
         return res.sendFile('error.svg', fileConfig);
--- a/test/api_test.js
+++ b/test/api_test.js
@@ -82,7 +82,7 @@ suite('api', () => {
 
   setup(async () => {
     github = await helper.load('github');
-    github.inst(9090).setRepositories('coolRepo', 'anotherCoolRepo', 'awesomeRepo');
+    github.inst(9090).setRepositories('coolRepo', 'anotherCoolRepo', 'awesomeRepo', 'nonTCGHRepo');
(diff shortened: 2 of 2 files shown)


owlishDeveloper (approved):

src/api.js
```diff
@@ -314,8 +314,13 @@ api.declare({
...
       let status = statuses.find(statusObject => statusObject.creator.id === taskclusterBot.id);
 
-      // Then we send a corresponding image.
-      return res.sendFile(status.state + '.svg', fileConfig);
+      if (status) {
+        // If we got a status, send a corresponding image.
+        return res.sendFile(status.state + '.svg', fileConfig);
+      } else {
+        // otherwise, huh, a commit without a status..
+        return res.sendFile('error.svg', fileConfig);
```
owlishDeveloper: the only thing, should it maybe be "no info yet" image?

test/api_test.js
```diff
@@ -82,7 +82,7 @@ suite('api', () => {
 
   setup(async () => {
     github = await helper.load('github');
-    github.inst(9090).setRepositories('coolRepo', 'anotherCoolRepo', 'awesomeRepo');
+    github.inst(9090).setRepositories('coolRepo', 'anotherCoolRepo', 'awesomeRepo', 'nonTCGHRepo');
```
owlishDeveloper, line 85: (a nit: it's not necessarily non tc-gh repo. As in tc-cli case, when tc is not used for pushes)

src/api.js
```diff
@@ -314,8 +314,13 @@ api.declare({
...
+      }
     } catch (e) {
       if (e.code < 500) {
         return res.sendFile('error.svg', fileConfig);
```
owlishDeveloper: I also thought: what if we just change `catch` to throw errors whose code is >= 500, and send error image for the rest? That TypeError gets caught here, does'n it?

djmitche: I think that might be too broad -- I'd like to get alerted in Sentry to programming errors in this try block.

test/api_test.js
djmitche: I meant "nonTCStatusInThisRepo".. just shortened.  But, good point :)

src/api.js
djmitche: is that "newrepo.svg"?

owlishDeveloper: yeah

src/api.js
owlishDeveloper: oh, ok

djmitche:
updated to use `norepo.svg`

PR #180, opened 2017-03-13
Update api_test.js

owlishDeveloper:
Update the test for the recent change ("newrepo" image instead of "error" should be returned in that test)


djmitche (approved):
Thanks :embarassed:

PR #181, opened 2017-03-22
[WIP] "InstalledFor" doesn't work for taskcluster org. Solution v.01

owlishDeveloper:
This is to try and remedy "Installed for" API method. I am trying to glue all the pages together. I'm not sure if it's quite this simple... For a start I want to see if I use hasNextPage() and getNextPage() correctly.

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -354,13 +354,27 @@ api.declare({
     try {
       let instGithub = await this.github.getInstallationGithub(ownerInfo.installationId);
       let reposList = await instGithub.integrations.getInstallationRepositories({});
+      debug(`REPOSLIST: ${JSON.stringify(reposList)}`);
 
       // GitHub API returns an array of objects, each of wich has an array of repos
-      let installed = reposList.repositories.map(repo => repo.name).indexOf(repo);
+      var installed = reposList.repositories.map(repo => repo.name).indexOf(repo);
+      debug(`INSTALLED var: ${installed}`);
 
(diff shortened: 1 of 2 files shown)


src/api.js
```diff
@@ -354,10 +354,19 @@ api.declare({
...
       let instGithub = await this.github.getInstallationGithub(ownerInfo.installationId);
       let reposList = await instGithub.integrations.getInstallationRepositories({});
+      debug(`REPOSLIST: ${JSON.stringify(reposList)}`);
 
       // GitHub API returns an array of objects, each of wich has an array of repos
       let installed = reposList.repositories.map(repo => repo.name).indexOf(repo);
 
+      if (installed === -1 && instGithub.hasNextPage()) {
+        while (installed === -1 && instGithub.hasNextPage()) {
+          reposList = instGithub.getNextPage();
```
djmitche: I think you need an `await` here!

owlishDeveloper: right!

owlishDeveloper: updated that, thanks!

owlishDeveloper:
I corrected the logic, now it should work properly. However, before merging, I'd deploy the thing to regular tc-gh (not staging), to check the pagination stuff. I leave all the debugging statements for now. (I'll be away the first half of Monday. I'll ping you, @imbstack, on IRC somewhere in the afternoon so we could test it?)

imbstack:
This work will continue in #188.

PR #184, opened 2017-04-10
Bug 1337820: update tests to use fake PulseConnection

djmitche:
This is a relatively small change -- it just looks big because I changed the indentation when I removed `noConnect`.  Instead, we set up the connection as usual, but config.yml says it's a fake connection.  Aside from that, this is basically changing `.emit` to `.fakeMessage` and adding a few fields.


owlishDeveloper:
If the changes in yarn.lock are OK, I approve!

owlishDeveloper (approved):

PR #174, opened 2017-03-07
[WIP] API to make badges clickable

owlishDeveloper:
First draft of the API function. Not sure how to test this yet, just wanted to show this and get any feedback :)

PR #189, opened 2017-05-05
Added API to make the badges clickable

owlishDeveloper:
Continued work from #174 

Added the new API, which uses Express's redirect() method to redirect the user to the link contained in the status object returned by GitHub.

My three API functions have some duplicate code, so I put that into two separate functions (I haven't touched the other two APIs here yet - I think I'll better do that in a separate PR later).

I also devised a test for the API. It doesn't test much yet, and it's weird, because I'm not sure where to redirect the darn thing (as of now, I redirect to google.com, but I think it's wrong), so any advice there would be much appreciated!

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -376,3 +376,61 @@ api.declare({
   }
   return res.reply({installed: false});
 });
+
+// Below is the draft code to make the badges clickable.
+// In the finished thing, the helper function will go to the beginning of the module,
+// and will be used in other APIs as well.
+api.declare({
+  name: 'taskLink',
+  title: 'Redirects to the task inspector page',
+  description: [
+    'Builds a link to the task inspector for the given task group',
+    ' and redirects the user to that page.',
+  ].join('\n'),
+  stability: 'experimental',
+  method: 'get',
+  route: '/taskLink/:owner/:repo/:branch',
+}, async function(req, res) {
+  // Extract owner, repo and branch from request into variables
+  let {owner, repo, branch} = req.params;
+
+  let authenticate = installationAuthenticate.bind(this);
+  let instGithub = await authenticate(owner);
+
+  // Get task group ID
+  if (instGithub) {
+    try {
+      let findStatus = findTCstatus.bind(this);
+      let status = await findStatus(instGithub, owner, repo, branch);
+
+      return res.redirect(status.target_url);
+    } catch (e) {
+      debug(`Error creating link: ${JSON.stringify(e)}`);
+      return res.status(500).send();
+    }
+  }
+  return res.status(404).send();
+});
+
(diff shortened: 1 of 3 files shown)


imbstack:
This is looking really good. I think we should rename some of the api functions before we ship, for real, but maybe that can happen in another PR as well. I think the right thing to do instead of the `google.com` hack is to tell requests not to follow redirects (look up followRedirect [here](https://github.com/request/request#requestoptions-callback)). 

Nice work :)

owlishDeveloper:
Thank you Brian! I'll look that followRedirect thing. Which names would you suggest for the functions?

imbstack:
Let's figure out api design stuff in a separate PR. I want to poke at it for a bit once we land this and see what looks nice.

src/api.js
```diff
@@ -376,3 +376,61 @@ api.declare({
...
+
+// Below is the draft code to make the badges clickable.
+// In the finished thing, the helper function will go to the beginning of the module,
+// and will be used in other APIs as well.
+api.declare({
+  name: 'taskLink',
+  title: 'Redirects to the task inspector page',
+  description: [
+    'Builds a link to the task inspector for the given task group',
+    ' and redirects the user to that page.',
```
imbstack, line 388: You don't need that extra space at the beginning of this line.

imbstack (approved):
r+

With nit and once the redirect stuff is updated :)

src/api.js
djmitche: Also, that's not really what this method is doing anymore, since it's just returning status.target_url.

src/api.js
```diff
@@ -376,3 +376,61 @@ api.declare({
...
+  // Extract owner, repo and branch from request into variables
+  let {owner, repo, branch} = req.params;
+
+  let authenticate = installationAuthenticate.bind(this);
+  let instGithub = await authenticate(owner);
+
+  // Get task group ID
+  if (instGithub) {
+    try {
+      let findStatus = findTCstatus.bind(this);
```
djmitche, line 403: Rather than bind'ing to a function, I'd prefer to just pass the context explicitly:

```
await findStatus(instGithub, owner, repo, branch, this)
```
...
```
async function findTCstatus(github, owner, repo, branch, context) {
... context.cfg.app.botName ...
}
```
(same for `installationAuthenticate`, above)

djmitche (commented):
Maybe those are the API changes @imbstack is talking about?

src/api.js
owlishDeveloper: OK, I'll do that!

src/api.js
owlishDeveloper: Both done

owlishDeveloper:
Thank you guys!

djmitche (approved):

PR #193, opened 2017-05-09
Update public api

imbstack:
This is my idea for rearranging our http endpoints and library function names a bit to align more with the resource mapping we can have in addition to general flexibility.

Let me know what you think!

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -284,12 +284,12 @@ api.declare({
   name: 'badge',
   title: 'Latest Build Status Badge',
   description: [
-    'Checks the status of the latest build of a given branch ',
-    'and returns corresponding badge image.',
+    'Checks the status of the latest build of a given branch',
+    'and returns corresponding badge svg.',
   ].join('\n'),
   stability: 'experimental',
   method: 'get',
-  route: '/badge/:owner/:repo/:branch',
+  route: '/repository/:owner/:repo/:branch/badge.svg',
 }, async function(req, res) {
   // Extract owner, repo and branch from request into variables
   let {owner, repo, branch} = req.params;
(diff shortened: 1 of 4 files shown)


src/api.js
```diff
@@ -377,20 +377,19 @@ api.declare({
...
+    'For a given branch of a repository, this will always point',
+    'to a status page for the most recent task triggered by that',
+    'branch.',
+    '',
+    'Note: This may be a redirect rather than a direct link.',
   ].join('\n'),
   stability: 'experimental',
   method: 'get',
-  route: '/taskLink/:owner/:repo/:branch',
+  route: '/repository/:owner/:repo/:branch',
```
owlishDeveloper: It will not be confusing, will it? From such a general name, I guess we plan to make more APIs that have `repository/` in their route?

src/api.js
```diff
@@ -377,20 +377,19 @@ api.declare({
...
+  title: 'Status Page for Branch',
   description: [
-    'Finds a link to the task inspector for the given task group',
-    'in the status object returned by GitHub,',
-    'and returns the link to redirect the user to that page.',
+    'For a given branch of a repository, this will always point',
+    'to a status page for the most recent task triggered by that',
+    'branch.',
+    '',
+    'Note: This may be a redirect rather than a direct link.',
```
djmitche: Isn't it always a redirect?

src/api.js
djmitche: All of the API methods start with `/repository` :)

But you're right - it seems like it should be `/repository/:owner/:repo/:branch/latest` or something like that, to parallel the `badge.svg` route.

djmitche (changes requested):
Minor changes, but thanks for this cleanup!

owlishDeveloper (approved):

djmitche (approved):

PR #197, opened 2017-05-17
Bug 1343871 - Supports posting custom status to changesets. r=dustin

ochameau:
https://bugzilla.mozilla.org/show_bug.cgi?id=1343871

I simply mapped Github API to tc-github:
https://developer.github.com/v3/repos/statuses/#create-a-status

I added a test, but I've no idea if that actually works!
Otherwise I think I just miss a scope check before proceeding with this patch?

### DIFF
--- a/schemas/create-status.yml
+++ b/schemas/create-status.yml
@@ -1,0 +1,23 @@
+$schema:  http://json-schema.org/draft-04/schema#
+title:        "Create status"
+description: |
+  Create a commit status on GitHub.
+  Full specification on [GitHub docs](https://developer.github.com/v3/repos/statuses/#create-a-status)
+type:         object
+properties:
+  state:
+    type: string
+    enum: ['pending', 'success', 'error', 'failure']
+    description: The state of the status.
+  target_url:
+    type: string
+    description: The target URL to associate with this status. This URL will be linked from the GitHub UI to allow users to easily see the 'source' of the Status.
+  description:
+    type: string
+    description: A short description of the status.
+  context:
+    type: string
+    description: A string label to differentiate this status from the status of other systems.
+additionalProperties: false
+required:
+  - state
--- a/src/api.js
+++ b/src/api.js
@@ -479,3 +479,44 @@ api.declare({
 
   return res.status(404).send();
 });
+
+api.declare({
+  name: 'createComment',
+  title: 'Post a comment on a given GitHub Issue or Pull Request',
+  description: [
+    'For a given Issue or Pull Request of a repository, this will write a new message.',
+  ].join('\n'),
+  stability: 'experimental',
+  method: 'post',
+  // route and input (schema) matches github API
+  // https://developer.github.com/v3/issues/comments/#create-a-comment
+  // number is a Issue or Pull request ID. Both share the same IDs set.
+  route: '/repository/:owner/:repo/issues/:number/comments',
+  input: 'create-comment.json',
+  scopes: [['github:create-comment:${owner}/${repo}']],
+}, async function(req, res) {
+  // Extract owner, repo and number from request into variables
+  let {owner, repo, number} = req.params;
+  // Extract body from POST attributes
+  let {body} = req.body;
+
+  let instGithub = await installationAuthenticate(owner, this.OwnersDirectory, this.github);
+
+  if (instGithub) {
+    try {
+      await instGithub.repos.createComment({
+        owner,
+        repo,
+        number,
+        body,
+      });
+
+      return resolve(res, 200, 'Comment created!');
+    } catch (e) {
+      debug(`Error creating status: ${JSON.stringify(e)}`);
+      return res.status(500).send();
+    }
+  }
+
+  return res.status(404).send();
+});
--- a/src/api.js
+++ b/src/api.js
@@ -433,3 +433,49 @@ api.declare({
   }
   return res.status(404).send();
 });
+
+api.declare({
+  name: 'createStatus',
+  title: 'Post a status against a given changeset',
+  description: [
+    'For a given changeset (SHA) of a repository, this will attach a "commit status"',
+    'on github. These statuses are links displayed next to each revision.',
+    'The status is either OK (green check) or FAILURE (red cross), ',
+    'made of a custom title and link.',
+  ].join('\n'),
+  stability: 'experimental',
+  method: 'post',
+  // route and input (schema) matches github API
+  // https://developer.github.com/v3/repos/statuses/#create-a-status
+  route: '/repository/:owner/:repo/statuses/:sha',
+  input: 'create-status.json',
+  scopes: [['github:create-status:${owner}/${repo}']],
+}, async function(req, res) {
+  // Extract owner, repo and sha from request into variables
+  let {owner, repo, sha} = req.params;
+  // Extract other attributes from POST attributes
+  let {state, target_url, description, context} = req.body;
+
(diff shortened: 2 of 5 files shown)


djmitche:
You've anticipated the review well :)

GET requests should not have side effects, so I think this should be a POST, and most of the parameters given in the query should, instead, be passed in a JSON body.

The scopes are a bit of a trickier question.  I'll be interested to hear @imbstack's ideas here, but I can imagine a case where taskcluster-github gives each task it creates a scope like `github:update-status:<org>/<repo>/<sha>`, and then requires that scope in order to update a status.  That would cover most cases, and would also allow us to permit the weirder cases where, for example, repo A wants to update the status for repo B (grant repo B `github:update-status:org/repo-b/*`).

That might deserve an [RFC](https://github.com/taskcluster/taskcluster-rfcs/) if it's sufficiently complicated.

ochameau:
@djmitche I pushed a new revision, but really I can't understand how to make that work around POST instead of GET. I wasn't able to find any example, not in taskcluster-github, nor taskcluster-lib-api.
githubWebHookConsumer uses POST but dot not have any schema. It isn't clear how you would define argument coming from the routes and the one from the POST body.

Otherwise, how do I add a scope check? Are they stored somewhere? Would it be on this.cfg.scopes?

djmitche:
https://github.com/taskcluster/taskcluster-secrets might be a better example.  TC-github is a little weird since it gets called by Github which doesn't really know anything about TC APIs.

Putting a secret
  https://github.com/taskcluster/taskcluster-secrets/blob/master/src/api.js#L31
uses `secret.json` which maps (switching the extension for reasons I don't really understand) to
  https://github.com/taskcluster/taskcluster-secrets/blob/master/schemas/secret.yml
which is a JSON schema object.

Hopefully that helps?

ochameau:
Thanks @djmitche, this pull request is green again and accomodate your comments about using POST and using scopes. 
Is putting `scopes: [['github:create-status:<owner>/<repo>/<sha>']],` in `api.declare` call enough?
Then I imagine you would set this scope for tc-github tasks from somewhere else?

schemas/create-status.yml
```diff
@@ -0,0 +1,22 @@
+$schema:  http://json-schema.org/draft-04/schema#
+title:        "Create status"
+description: |
+  Create a commit status on GitHub
+type:         object
+properties:
+  state:
+    type: string
+    enum: ['pending', 'success', 'error', 'failure']
+    description: github status associated with the build.
```
imbstack: Let's make the descriptions full sentences starting with capital letter and ending with period. Not sure we do that correctly everywhere already, but may as well do it here!

imbstack: Maybe a link to where github defines these as well? This is rendered as markdown on our docs site so we can use that link syntax.

src/api.js
```diff
@@ -433,3 +433,50 @@ api.declare({
...
+    'The status is either OK (green check) or FAILURE (red cross), ',
+    'made of a custom title and link.',
+  ].join('\n'),
+  stability: 'experimental',
+  method: 'post',
+  // route and query matches github API
+  // https://developer.github.com/v3/repos/statuses/#create-a-status
+  route: '/repository/:owner/:repo/statuses/:sha',
+  input: 'create-status.json',
+  scopes: [['github:create-status:<owner>/<repo>/<sha>']],
```
imbstack: We don't really want scopes based on sha, right? Branch makes a bit more sense in this context. That way we can control whether or not you can set scopes on master, but a sha-based scopes is only useful once.

schemas/create-status.yml
```diff
@@ -0,0 +1,22 @@
...
+    description: link to put as github status.
+  description:
+    type: string
+    description: title displayed on the github status.
+  context:
+    type: string
+    description: string label to differenciate github statuses.
+additionalProperties: false
+required:
+  - state
```
imbstack: Afaict we don't set any defaults for the non-required classes here. Is that expected and is context/description/target_url allowed to be null in the github api?

imbstack (commented):
Looks really good so far!

schemas/create-status.yml
ochameau: Yes, state is the only mandatory attribute.
I'm defaulting context to "default" to match their API behavior.

src/api.js
ochameau: Yes, it may be too precise. Actually, it would prevent posting status for all but lastest changeset of a pull request.
Having said that, it isn't clear how I determine on which branch the current task is running against? And it is even more mysterious how could I know that before calling `api.deplare()`?
(As we have to know that to set the `scopes` attribute right here.

ochameau: Ok. So using `deferAuth: true` and `req.satisfies` works fine.
But the issue is that GitHub API doesn't allow getting the branch name of a given commit :/
http://stackoverflow.com/questions/29992441/github-api-identify-the-branches-that-contain-a-given-commit
I also looked at all available `Repositories` API from GitHub:
https://developer.github.com/v3/repos/
As well as Git Data:
https://developer.github.com/v3/git/
But haven't found any that exposes the branch.

I thought we could assert Task branch we are running but that's not what we want to assert, it would still allow passing any SHA from any branch/repo.

The other assertion I could think about would be the SHA for "push" tasks and the "Commit range" for "pull request" tasks. Do we have this information somewhere? Having said that I'm not sure it will be easy to match commit ranges from a scope string.

Note that the scope definition should be easier for Pull request comments API as we would have something like `github:update-status:<org>/<repo>/<PR#ID>`.

Otherwise, is this function the one to modify to correctly flag the tasks with the new necessary scopes?
https://github.com/taskcluster/taskcluster-github/blob/master/src/intree.js#L14
Or does it lives somewhere else in another project?

ochameau:
Just pushed another commit to support pull request comments.
The scope story looks easier `github:create-comment:${owner}/${repo}/${number}`. (number is the PR ID)
Would you like me to open another pull request and move forward on comments first?

imbstack:
We can do it all at once! I think this afternoon I'll have time to look at the scopes stuff with this.

ochameau:
Note that I got to ask Github about this limitation (getting the branches of a changeset). And they confirmed it, with what looks like slow workarounds:
http://stackoverflow.com/questions/23899329/list-of-branches-a-commit-appears-on/23970412#23970412

imbstack (commented):
I think the easiest thing for us to do is just make the scopes for `createStatus` only on `owner` and `repo`. We're not going to restrict to a `sha` ever afaict. The same goes for `createComment` and `number`. Does that make any sense? I think if we just remove them we can land this.

ochameau:
Yes, it sounds find to check only `owner` and `repo`. Posting status and comment should not allow much escalation. It is mostly a possibility to spam/pollute things. Restricting to only the pull request sounds more like a nice to have and very perfect limitation.

But this current patch is already so much better than what anyone do on other CI! 
Without this, you end up using GITHUB_TOKEN via secrets/private env variable, which always comes with way too many privileges...

ochameau:
I updated the patch with the new scopes.
Again, I'm still wondering where these scopes are going to be set?
Will it be done manually?

imbstack (approved):

imbstack:
@ochameau the task will automatically have the appropriate scopes if [taskclusterProxy](https://docs.taskcluster.net/reference/workers/docker-worker/docs/features#feature-taskclusterproxy-) is used from within the task. I think this is good to land now.

@djmitche and I think we can make the scopes more specific later, but it will take some work first, so I'm going to track that work in taskcluster/taskcluster-github#198

I'll land this soon.

PR #203, opened 2017-06-22
Stop warning users they cannot run on PR if request not a PR.

imbstack:
We received a report from a user that they were getting the "you cannot PR!" warning on all commits to a repo. An [example they sent](https://github.com/mozilla-bteam/bmo/commit/38614185c018d29ae61f94e822480b0ea572155d#commitcomment-22590197) shows this to be true and investigating their [.taskcluster.yml](https://github.com/mozilla-bteam/bmo/blob/master/.taskcluster.yml) we can see that they don't want to do anything for PR anyway. I believe what's happening is that we check to see if this can run on a PR before we determine if they've asked for a run on a PR. This patch just moves the check for `allowPullRequests` to after we determine that this is actually being run on a `pull_request.*` and not a push.

### DIFF
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -284,6 +284,28 @@ async function jobHandler(message) {
     throw e;
   }
 
+  let groupState = 'pending';
+  let taskGroupId = 'nonexistent';
+  let graphConfig;
+
+  // Now we can try processing the config and kicking off a task.
+  try {
+    graphConfig = this.intree({
+      config: repoconf,
+      payload: message.payload,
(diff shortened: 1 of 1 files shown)


floatingatoll:
cc @dylanwh :)

djmitche (approved):
This code is getting pretty hairy :)

floatingatoll:
I believe an interesting side effect here is that PRs still show orangle-circle for Taskcluster.

https://github.com/mozilla-bteam/bmo/pull/120

PR #205, opened 2017-07-20
update docs for v0 .taskcluster.yml

djmitche:
Mostly just moving usage.md, reorganizing its content, and adding a bit of context (like, where do you put `events`?)

### DIFF
--- a/docs/taskcluster-yml-v0.md
+++ b/docs/taskcluster-yml-v0.md
@@ -1,43 +1,56 @@
 ---
-title: Using Taskcluster for Github Projects
-order: 40
+title: .taskcluster.yml
+order: 20
 ---
 
-Taskcluster is easy to set up for simple CI cases and very expressive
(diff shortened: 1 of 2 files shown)


docs/taskcluster-yml-v0.md
```diff
@@ -54,31 +67,54 @@ You can modify a task definition so that it will only run for specific GitHub ev
...
         branches:
           - master
 ```
+
 Branch filtering doesn't work for releases.
 
----
+## Who Can Trigger Jobs?
+
+TaskCluster-Github will always build pushes and releases.  For pull requests, two policies are available:
```
imbstack: Check out that uppercase C :p

imbstack (approved):

docs/taskcluster-yml-v0.md
djmitche: Woah!  Where did that come from!

PR #210, opened 2017-08-11
Add 'queue:scheduler-id:taskcluster-github'

jonasfj:
Might need to configurable...


imbstack:
r- needs to be configurable

imbstack:
We added this scope to `repo:github.com/*` instead.

PR #213, opened 2017-09-27
Allow interactions from people who have set non-emails as their email

imbstack:
This is not the prettiest fix in the world, but tc-gh is already not very pretty. This should allow people with their github profile emails set to things like "foo[at]bar[dot]com" to use tc-gh.

[Bug 1403537](https://bugzilla.mozilla.org/show_bug.cgi?id=1403537)

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "url": "https://github.com/taskcluster/taskcluster-github.git"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


djmitche (approved):
Nicely done!

@alexandrasp: FYI, hopefully this will fix the issus you're seeing!

PR #215, opened 2017-11-13
Bug 1416856 - Ban non-ascii branch names

imbstack:
This is a bit ugly, but as we've discussed before, tc-gh is already an ugly pile of hacks :/

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -18,7 +18,7 @@ tasks:
           - push
     payload:
       maxRunTime: 3600
-      image: "node:6"
+      image: "node:8"
       env:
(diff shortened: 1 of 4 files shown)


djmitche (approved):
One way to look at the issue is that tc-github can't invent a valid scope for a branch with unicode characters.  This change should probably have some documentation, and that might be a good angle to take in the docs (rather than "we don't support Unicode" which sounds like we're still running Windows 95..)

There's room in the future to support unicode by encoding the branch name or by using some fixed string like `unencodable`, but we could take a little longer to think about that first.

It's interesting that this error got all the way through to treeherder -- I would have expected the queue to fail when it tried to authorize createTask with authorizedScopes containing invalid scopes.

imbstack:
It did fail at `createTask` too. I'm honestly not sure how it ended up in treeherder-land.

PR #221, opened 2017-12-04
Bug 1422388 - Add some checks for task group status before updating github status

imbstack:
[Bug context](https://bugzilla.mozilla.org/show_bug.cgi?id=1422388)

### DIFF
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -166,6 +166,23 @@ async function statusHandler(message) {
   debug(`Handling state change for task-group ${taskGroupId}`);
 
   let state = 'success';
+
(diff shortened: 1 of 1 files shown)


djmitche (approved):
This will mean that retriggering a task in a taskgroup won't make the taskgroup go green.  I *think* that's already the case, though, right?

imbstack:
We've never really had a concrete story for how retriggers interact with tc-gh.  I think with the current implementation, retriggering will make the taskgroup go green if the retriggered task passed regardless of if there are any other failed tasks in the group. I think the current behavior is actually kinda broken in a way too.

We need to think about that whole side of things more I expect.

PR #222, opened 2018-01-10
Bug 1422308 - Taskcluster bot replies to PRs with different subject

ydidwania:
I added a method `pullRequests.createComment` to [tests/github-auth.js](https://github.com/taskcluster/taskcluster-github/pull/222/files#diff-99979e9c6c4c7ef22f28ff0ad50cae77R54) for now and used it to comment on a PR in [src/handlers.js](https://github.com/taskcluster/taskcluster-github/pull/222/files#diff-17e95b0c17bfcf5e5d36214802cc8b06R349) . For this I had to add `event.number` to the mock message for pull requests in [test/handlers_test.js](https://github.com/taskcluster/taskcluster-github/pull/222/files#diff-bc6ee6e86e4196af3b4ce47133f58b91R74).

### DIFF
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -125,7 +125,7 @@ class Handlers {
   }
 
   // Send an exception to Github in the form of a comment.
-  async createExceptionComment({instGithub, organization, repository, sha, error, number}) {
+  async createExceptionComment({instGithub, organization, repository, sha, error, pullNumber}) {
     let errorBody = error.body && error.body.error || error.message;
     // Let's prettify any objects
     if (typeof errorBody == 'object') {
@@ -143,11 +143,11 @@ class Handlers {
     // Warn the user know that there was a problem handling their request
     // by posting a comment; this error is then considered handled and not
     // reported to the taskcluster team or retried
-    if (number) {
+    if (pullNumber) {
       await instGithub.pullRequests.createComment({
         owner: organization,
         repo: repository,
-        number,
+        pullNumber,
         body,
       });
       return;
@@ -258,7 +258,7 @@ async function jobHandler(message) {
   let repository = message.payload.repository;
   let sha = message.payload.details['event.head.sha'];
   //console.log('..message passed to jobHandler', message);
-  let number = message.payload.details['event.pullNumber'];
+  let pullNumber = message.payload.details['event.pullNumber'];
   //console.log('..pr number', number);
   if (!sha) {
     debug('Trying to get commit info in job handler...');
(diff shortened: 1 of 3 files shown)


ydidwania:
I have a lot of `console.log` statements now . I will remove them once you approve this PR :)

imbstack:
This is looking good so far! Only issues I've seen so far:

The PR objects don't have a `number` field in tc-gh, but rather a `pullNumber`. You can see where the object is created [here](https://github.com/taskcluster/taskcluster-github/blob/master/src/api.js#L39).

In addition, I believe there are more `createCommitComment` calls that we could try to replace here.

Now that I say that we also have one more issue to fix up here. It's not that we want to replace `createCommitComment` in every case, only when we are responding to a PR. We will need to wrap the call to the function in some sort of check to decide which one to call.

ydidwania:
I modified `createExceptionComment` to use `pullRequests.createComment()` when `number` is defined, and  `repos.createCommitComment()` otherwise. I think this will ensure we are always responding to a PR in the correct way.

src/handlers.js
```diff
@@ -339,21 +354,23 @@ async function jobHandler(message) {
...
+          number,
           body,
         });
         return;
       }
     } catch (e) {
       if (e.name === 'YAMLException') {
         let docsLink = 'https://docs.taskcluster.net/reference/integrations/github/docs/usage#who-can-trigger-jobs';
-        await instGithub.repos.createCommitComment({
+        await instGithub.pullRequests.createComment({
```
imbstack, line 370: We only want to use `createComment` instead of `createCommitComment` if this is a pull request.

src/handlers.js
```diff
@@ -339,21 +354,23 @@ async function jobHandler(message) {
           '```\n',
           '</details>',
         ].join('\n');
-        await instGithub.repos.createCommitComment({
+        //console.log(instGithub.pullRequests);
+        //console.log('..pr number' organization);
+        await instGithub.pullRequests.createComment({
```
imbstack, line 359: We only want to use `createComment` instead of `createCommitComment` if this is a pull request.

src/handlers.js
```diff
@@ -245,6 +257,9 @@ async function jobHandler(message) {
   let organization = message.payload.organization;
   let repository = message.payload.repository;
   let sha = message.payload.details['event.head.sha'];
+  //console.log('..message passed to jobHandler', message);
+  let number = message.payload.details['event.pullNumber'];
```
imbstack: Let's call this `pullNumber` instead of just `number` to make it more obvious what it is!

imbstack (commented):
Looking very good! Almost ready.

src/handlers.js
ydidwania: This  (and the next call) is enclosed by the `if (message.payload.details['event.type'].startsWith('pull_request.'))` check on L341 above , so these are executed only in case of pull requests

imbstack (approved):
This looks great! Thanks very much. I'll try this out today.

imbstack:
Oh, I forgot we were going to remove the `console.log`s first. I'll go do that now.

ydidwania:
Okay. Thanks for all the help on this! :)

PR #223, opened 2018-01-19
Bug 1431750 - add additionalProperties to schemas where missing

petemoore:
Thanks Brian!


imbstack (approved):
From reading the bug it sounds like maybe we don't want to change this when there is no `properties` field? Otherwise this looks good.

petemoore:
Thanks Brian - yes, we can close this now, indeed. Thanks for looking into it, and sorry for the noise!

PR #224, opened 2018-01-23
Update to scope expressions

imbstack:


djmitche (approved):
I gather from my inbox that this got tested in staging :)

PR #225, opened 2018-01-30
Bug 1344912 - Support Tag events

ydidwania:
Aim is to provide an option to the owner of the repo in  `.taskcluter.yml` to schedule a task when only a tag is pushed by adding `tag` to `tasks.extra.github.events`


docs/taskcluster-yml-v0.md
```diff
@@ -63,7 +63,7 @@ The available events are:
   * `pull_request.synchronize` (a new commit is pushed to the branch in the PR. NOTE: There is no 'd' at the end of 'synchronize')
   * `push`                     (a push is made directly to the repo)
   * `release`                  (a new release published in any branch of the repo. NOTE: New tags are actually push events in github)
-
+  * `tag`
```
imbstack: Let's add a comment to this as with the other events.

src/api.js
```diff
@@ -57,15 +58,21 @@ function getPushDetails(eventData) {
     'event.base.user.login': eventData.sender.login,
 
     'event.head.ref': ref,
-    'event.head.repo.branch': branch,
     'event.head.repo.name': eventData.repository.name,
     'event.head.repo.url': eventData.repository.clone_url,
     'event.head.sha': eventData.after,
     'event.head.user.login': eventData.sender.login,
-    'event.head.user.id': eventData.sender.id,
+    'event.head.user.id': eventData.sender.id,
```
imbstack: You have an extra space at the end of this line.

src/api.js
```diff
@@ -48,7 +48,8 @@ function getPushDetails(eventData) {
   // parsing the ref refs/heads/<branch-name> is the most reliable way
   // to get a branch name
   let branch = ref.split('/').slice(2).join('/');
-  return {
+  let isTagEvent = ref.split('/')[1] == 'tags';
```
imbstack: Let's go for `===` just for the sake of it 😆

imbstack (commented):
This is a great start! I'll expect you'll want to include  `event.head.tag` in [the env vars](https://github.com/taskcluster/taskcluster-github/blob/master/src/intree.js#L45) in addition to updating [the docs](https://github.com/taskcluster/taskcluster-github/blob/master/docs/taskcluster-yml-v0.md#token-substitution-and-environment-variables).

ydidwania:
I get the below error on [this](https://pastebin.mozilla.org/9077280) webhook I created for tag pushes while running the [webhook tests](https://github.com/taskcluster/taskcluster-github/blob/master/test/webhook_test.js). The function call  from [jsonHttpRequest](https://github.com/taskcluster/taskcluster-github/blob/master/test/helper.js#L32) doesn't return in a long time and hence `socket hang up`
```
  1) webhook
       Tag:
     Uncaught Error: socket hang up
      at createHangUpError (_http_client.js:343:15)
      at Socket.socketOnEnd (_http_client.js:435:23)
      at endReadableNT (_stream_readable.js:1045:12)
      at _combinedTickCallback (internal/process/next_tick.js:102:11)
      at process._tickDomainCallback (internal/process/next_tick.js:198:9)
```

ydidwania:
@imbstack , yes i'll update the docs and the env vars , once i am done with writing the tests.

test/helper.js
```diff
@@ -36,17 +36,17 @@ helper.jsonHttpRequest = function(jsonFile, options) {
...
   let jsonData = JSON.parse(fs.readFileSync(jsonFile));
   options.headers = jsonData.headers;
   return new Promise (function(accept, reject) {
     try {
       let req = http.request(options, accept);
       req.write(JSON.stringify(jsonData.body));
       req.end();
-    } catch (e) {
+      console.log('exiting helper');
+      console.log(accept, reject);
```
ydidwania: This `console.log` is executed but the control is still isn't returned to `statusTest` in tests/webhook_test.js` from where it is called.

test/data/webhooks/webhook.tag_push.json
```diff
@@ -0,0 +1,10 @@
+{"headers": {"Host": "requestb.in",
+"Accept": "*/*",
+"User-Agent": "GitHub-Hookshot/6f33334",
+"X-GitHub-Event": "push",
+"X-GitHub-Delivery": "7639a9808fq-11w3-2359-6266dd99y98d",
+"content-type": "application/json",
+"X-Hub-Signature": "sha1=14b935ba703e45e7c6d64f3cc7e961d85e499015",
```
ydidwania: I have to change `X-Hub-Signature` so that it passes [this](https://github.com/taskcluster/taskcluster-github/blob/master/src/api.js#L207) condition in src/api.js
```
    if (!webhookSecrets.some(webhookSecret => {
      let calculatedSignature = generateXHubSignature(webhookSecret,
        JSON.stringify(body));
      return compareSignatures(calculatedSignature, xHubSignature);
```

djmitche:
Is this ready for the next review?

ydidwania:
I had not updated the headers of the test webhook `test/data/webhooks/webhook.tag_push.json` , I simply copied them from another test webhook :stuck_out_tongue_closed_eyes: . But  now they are good and the webhook_tests are running successfully.
Dustin, I will add the other tests, and then ask for a review. Now, there is not much to look at.

src/intree.js
```diff
@@ -125,7 +129,8 @@ module.exports.setup = function(cfg) {
         let events = task.task.extra.github.events;
         let branches = task.task.extra.github.branches;
         return _.some(events, ev => payload.details['event.type'].startsWith(_.trimEnd(ev, '*'))) &&
-          (!branches || branches && _.includes(branches, payload.details['event.base.repo.branch']));
+          (!branches || branches && _.includes(branches, payload.details['event.base.repo.branch']) ||
+            payload.details['event.type']==='tag');
```
ydidwania, line 135: Tags are applied to commits , hence tags and branches are not directly related in github. 
`event.base.repo.branch = 'tagname' ` for a tag event. So for a config like below 
```
        events:
          - push
          - tag
        branches:
          - master
```
No task for a tag event will be scheduled. hence I modified the above  check to make sure that branches are ignored for tag events. 
[Sample Payload for a tag push](http://help.appveyor.com/discussions/questions/1050-can-you-skip-a-build-if-there-is-a-tag)

src/intree.js
```diff
@@ -36,6 +36,10 @@ function completeInTreeConfig(config, payload) {
     config.scopes = [
       `assume:repo:github.com/${ payload.organization }/${ payload.repository }:release`,
     ];
+  } else if (payload.details['event.type'] == 'tag') {
+    config.scopes = [
+      `assume:repo:github.com/${ payload.organization }/${ payload.repository }:release`,
```
ydidwania: We want to treat tag events identical to release events. So i thought it would be appropriate to give them the same scopes as a release event. Is it a long process to add new scopes , if we want to do it differently ?

test/helper.js
```diff
@@ -36,17 +36,15 @@ helper.jsonHttpRequest = function(jsonFile, options) {
     path: '/v1/github',
     method: 'POST',
   };
-
   options = _.defaultsDeep(options, defaultOptions);
-
```
ydidwania, line 41: I'll correct these in the subsequent commits :stuck_out_tongue:

ydidwania:
@djmitche , @imbstack I need some help here. I've made quite a few changes. If one of you could review them , I can go on to add the handler_test , which should more or less complete this PR .

I believe , setting `event.base.repo.branch = 'tagname' ` for tag events is misleading, should I go ahead and remove it only for tag events ?

src/intree.js
djmitche: It's as easy as just naming a different scope.  So this could be `:tag:<tagName>` instead of `:release`.

src/api.js
```diff
@@ -48,7 +48,8 @@ function getPushDetails(eventData) {
   // parsing the ref refs/heads/<branch-name> is the most reliable way
   // to get a branch name
   let branch = ref.split('/').slice(2).join('/');
-  return {
+  let isTagEvent = ref.split('/')[1] === 'tags';
```
djmitche: Let's add a comment here that tags are identified by a ref like `heads/tags/<tagname>`.

src/api.js
```diff
@@ -196,8 +203,10 @@ api.declare({
   }
 
   let webhookSecrets = this.cfg.webhook.secret;
-  let xHubSignature = req.headers['x-hub-signature'];
+  //console.log('................secrets ', webhookSecrets);
```
djmitche: remove debug prints :)

src/api.js
```diff
@@ -57,15 +58,21 @@ function getPushDetails(eventData) {
...
-    'event.type': 'push',
+    'event.type': isTagEvent? 'tag' : 'push',
   };
+  if (isTagEvent) {
+    details['event.head.tag'] = branch;
+  } else {
+    details['event.head.repo.branch'] = branch;
+  }
+  return details;
+
```
djmitche: 👍

src/api.js
```diff
@@ -48,7 +48,8 @@ function getPushDetails(eventData) {
   // parsing the ref refs/heads/<branch-name> is the most reliable way
   // to get a branch name
   let branch = ref.split('/').slice(2).join('/');
-  return {
+  let isTagEvent = ref.split('/')[1] === 'tags';
+  let details = {
     'event.base.ref': ref,
     'event.base.repo.branch': branch,
```
djmitche: I think we should leave this out for tags -- there really is no branch in this case!

src/api.js
```diff
@@ -48,7 +48,8 @@ function getPushDetails(eventData) {
   // parsing the ref refs/heads/<branch-name> is the most reliable way
   // to get a branch name
   let branch = ref.split('/').slice(2).join('/');
```
djmitche: Probably this variable should be renamed to `refName` or something like that

src/intree.js
djmitche: I think this is a good fix, but also not setting `event.base.repo.branch` will help.

djmitche (approved):
This looks good to me!  @imbstack what do you think?

imbstack (approved):
Nice added tests! I think once you fix up the nits dustin mentionde we can land this. It ended up being a bigger change than I expected!

ydidwania:
Yes @imbstack . Looking at it now, do I need  to even add a `handlers_test`  for `tag events` .  As of now, `tests/handlers_test` doesn't check for `release events` as well.

djmitche:
Yes, a handlers test would be good too.  We likely overlooked the missing tests for release events.  Just adding one test that contains a tag-related PushEvent should be enough!

ydidwania:
@djmitche , I think this is  good to go and I hope this is the final review :)
I found a small bug like [this](https://github.com/taskcluster/taskcluster-github/blob/master/test/intree_test.js#L91). The test title mentions `Pull` but `event.type : push` is passed .

ydidwania:
Also i found that `GITHUB_HEAD_REF` has  been removed from [env vars](https://github.com/taskcluster/taskcluster-github/blob/master/src/intree.js#L45) but the [docs](https://github.com/taskcluster/taskcluster-github/blob/master/docs/taskcluster-yml-v0.md#push-metadata) have not been updated. 
 Edit : It is present in the payload for the currently running [taskcluster test](https://tools.taskcluster.net/groups/VEuPtSv7RTeAX9YX8VERYQ/tasks/VEuPtSv7RTeAX9YX8VERYQ/details) for these commits. ?

djmitche:
I see `GITHUB_HEAD_REF` here:
https://github.com/taskcluster/taskcluster-github/blob/08155fffccfc1545e83878c13012cfcacf3285e1/src/intree.js#L63

But if `event.head.ref` is not defined, then that env var will not be defined, either -- maybe that's what you're seeing?  In any case, little bugs like this make great additional PR's.  No need to file a Bugzilla bug in that case.

ydidwania:
Silly me, It got removed in the last commit :disappointed:

djmitche (approved):
Looks good!

@imbstack how do you feel about landing this? Should we maybe wait until next week?  We have informally agreed to try not to deploy anything this week since we have had so many production outages lately..

imbstack:
Yeah, this seems great to me. We can land this next week. I want to stick to our "no deploying" plan for at least a couple of days.

PR #226, opened 2018-02-02
Report table metrics like other services

jonasfj:
Using `monitor.prefix('data.builds')` makes a lot of sense, since a different deployment of `taskcluster-github`, should pass a different `project` name to monitor... if it passes the same project name, then it should be the same data in tables, etc... I don't know what the staging site does... but if uses the same data, it should also use the same monitor project name, if it uses different data tables, it shouldn't use the same project name.

In testing we use a different project name, well, we mock monitor... but the concept is reasonable I hope :)


imbstack (approved):
staging has a different name so all should be well!

PR #229, opened 2018-02-20
Bug 1433506 - Add ability to exclude branches

ydidwania:
Now, one can exclude branches by putting it under `task.extra.github.exclude.branches`

```yaml
version: 0
tasks:
  - ...
    extra:
      github:
        events:
          - push
        exclude:
          branches:
            - master
```
does not run tasks for pushed to the master branch. 
I went with this schema because it allows to exclude events in the future, while being compatible with the current in use configs.
@imbstack

### DIFF
--- a/src/intree.js
+++ b/src/intree.js
@@ -131,13 +131,11 @@ module.exports.setup = function(cfg) {
         let events = task.task.extra.github.events;
         let branches = task.task.extra.github.branches;
         let branch = payload.details['event.base.repo.branch'];
-        excludes = task.task.extra.github.exclude;
-        if (excludes) {
-          excludes = task.task.extra.github.exclude.branches;
-        }
+        let excludeBranches = task.task.extra.github.excludeBranches;
+
         return _.some(events, ev => payload.details['event.type'].startsWith(_.trimEnd(ev, '*'))) &&
-          (!(branches || excludes) || 
-            branches && _.includes(branches, branch) || excludes && !_.includes(excludes, branch) ||
+          (!(branches || excludeBranches) || 
+            branches && _.includes(branches, branch) || excludeBranches && !_.includes(excludeBranches, branch) ||
             payload.details['event.type']==='tag');
       });
 
--- a/src/intree.js
+++ b/src/intree.js
@@ -130,8 +130,14 @@ module.exports.setup = function(cfg) {
 
         let events = task.task.extra.github.events;
         let branches = task.task.extra.github.branches;
+        let branch = payload.details['event.base.repo.branch'];
+        excludes = task.task.extra.github.exclude;
+        if (excludes) {
+          excludes = task.task.extra.github.exclude.branches;
+        }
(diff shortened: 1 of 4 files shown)


imbstack (approved):
This looks really good to me. I'm _thinking_ that maybe we want to do `exclude_branches` instead just to keep this from getting so deeply nested. I certainly hear the rational of allowing us to exclude things later, but am not sure it is worth it. We can always add an `exclude_events` thing if we want later too. @djmitche any thoughts?

djmitche:
If it was
```
include:
  branches:
    - foo
```
then the parallel exclude would make sense.  But since it's
```
branches:
  - foo
```
I think
```
exclude-branches;
  - foo
```
has better parallel formation.  But, to be honest, I don't much care because soon this will be a JSON-e conditional :)

imbstack:
@ydidwania I think this will need to be updated now that we landed your other branch, but otherwise once we make it `excludeBranches`, we're good!

ydidwania:
@imbstack  I changed it to  `excludeBranches`. I rebased the changes on master so that there are no merge conflicts.

src/intree.js
```diff
@@ -130,8 +130,12 @@ module.exports.setup = function(cfg) {
 
         let events = task.task.extra.github.events;
         let branches = task.task.extra.github.branches;
+        let branch = payload.details['event.base.repo.branch'];
+        let excludeBranches = task.task.extra.github.excludeBranches;
```
imbstack, line 134: Oh, one other thing. If `task.task.extra` doesn't exist this will throw an error!

test/intree_test.js
```diff
@@ -145,6 +145,16 @@ suite('intree config', () => {
       tasks: [],
     });
 
+  buildConfigTest(
```
imbstack, line 148: Really there are a couple cases we should test maybe?

1. The case you have here
2. An exclude overridden by an include

imbstack (approved):
Looks great! Sorry for just a couple more nits.

src/intree.js
ydidwania: [This](https://github.com/taskcluster/taskcluster-github/blob/master/src/handlers.js#L304) line in src/handlers.js  takes care of it.

test/intree_test.js
ydidwania: I have added it in the docs that include takes precedence over exclude

imbstack: jhford march 23 through april 2

imbstack: Agreed, but once we state something in the docs it is nice to assert it in tests as well, right?

ydidwania: Yes. I'll make the test right away :)

staktrace:
@imbstack ping, looks like this is waiting for re-review? (edit: or is it ready to merge?)

imbstack (approved):

imbstack:
Ah, sorry! Landing now.

PR #234, opened 2018-03-22
Fix logic of exclude branches

imbstack:
This adds some tests and changes the logic of exclude branches a bit. [Bug 1433506](https://bugzilla.mozilla.org/show_bug.cgi?id=1433506)

### DIFF
--- a/src/intree.js
+++ b/src/intree.js
@@ -128,15 +128,33 @@ module.exports.setup = function(cfg) {
           return false;
         }
 
+        let event = payload.details['event.type'];
         let events = task.task.extra.github.events;
-        let branches = task.task.extra.github.branches;
         let branch = payload.details['event.base.repo.branch'];
+        let includeBranches = task.task.extra.github.branches;
         let excludeBranches = task.task.extra.github.excludeBranches;
 
-        return _.some(events, ev => payload.details['event.type'].startsWith(_.trimEnd(ev, '*'))) &&
-          (!(branches || excludeBranches) || 
-            branches && _.includes(branches, branch) || excludeBranches && !_.includes(excludeBranches, branch) ||
-            payload.details['event.type']==='tag');
+        if (includeBranches && excludeBranches) {
+          throw new Error('Cannot specify both `branches` and `excludeBranches` in the same task!');
+        }
+
+        return _.some(events, ev => {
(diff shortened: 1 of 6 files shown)


src/intree.js
```diff
@@ -128,15 +128,38 @@ module.exports.setup = function(cfg) {
           return false;
         }
 
+        let event = payload.details['event.type'];
         let events = task.task.extra.github.events;
         let branches = task.task.extra.github.branches;
         let branch = payload.details['event.base.repo.branch'];
```
djmitche: can we call this `includeBranches` for parallelism

src/intree.js
```diff
@@ -128,15 +128,38 @@ module.exports.setup = function(cfg) {
...
+
+          let included = branches && _.includes(branches, branch);
+          let excluded = excludeBranches && _.includes(excludeBranches, branch);
+
+          if (included && excluded) {
+            throw new Error(`Cannot include and exclude same branch. '${branch}' is in both lists!`);
+          }
+
+          if (branches && excludeBranches) {
+            return included && !excluded;
```
djmitche: isn't this option not allowed now?

djmitche (changes requested):

staktrace:
@djmitche Looks like the patch was updated, could you re-review? Thanks!

src/intree.js
```diff
@@ -128,15 +128,36 @@ module.exports.setup = function(cfg) {
...
+            return false;
+          }
+
+          if (event !== 'push') {
+            return true;
+          }
+
+          if (includeBranches && excludeBranches) {
+            throw new Error(`Cannot include and exclude same branch. '${branch}' is in both lists!`);
+          }
```
djmitche: I think the error message here is incorrect.  The condition here is that there is both a `branches` and an `excludeBranches`.  The README suggests this situation is allowed (although I don't know what that would mean -- if a branch isn't on either list, is it included or excluded?).

I think when we talked about this briefly, we decided that this case is not allowed -- you must use no more than one of `branches` or `excludeBranches`.  Then there's no concern with the same branch appearing in both lists.

djmitche (changes requested):

djmitche (approved):

PR #239, opened 2018-04-08
[WIP] jsonEfication

owlishDeveloper:
https://bugzilla.mozilla.org/show_bug.cgi?id=1335915

I decided to start with a WIP PR as opposed to asking questions. This is how I currently imagine the place of the `json-e` library in life, Universe and everything. I hope that's correct? What puzzles me is passing tests 😄

I guess I also need to add some tests for the version 1 YAML.


src/intree.js
```diff
@@ -103,14 +104,15 @@ module.exports.setup = function(cfg) {
...
-    }));
+    // config = jparam(config, _.merge(payload.details, {
+    //   $fromNow: (text) => tc.fromNowJSON(text),
+    //   timestamp: Math.floor(new Date()),
+    //   organization: payload.organization,
+    //   repository: payload.repository,
+    //   'taskcluster.docker.provisionerId': cfg.intree.provisionerId,
+    //   'taskcluster.docker.workerType': cfg.intree.workerType,
+    // }));
+    config = jsone(config);
```
djmitche: jsone takes two inputs -- template and context.  What's in that context is the hard part.  You can pattern it after what we do in mozilla-taskcluster -- https://docs.taskcluster.net/reference/integrations/mozilla-taskcluster/docs/taskcluster-yml

Note that we will also need to render the template on the master branch, without any pushes, to look for `allowPullRequests` and other security-related configuration.  Maybe that can have a different `tasks_for`, perhaps `access_check`?

src/intree.js
```diff
@@ -89,7 +90,7 @@ function completeInTreeConfig(config, payload) {
 module.exports.setup = function(cfg) {
   return function({config, payload, validator, schema}) {
     config = yaml.safeLoad(config);
-    let errors = validator(config, schema);
+    let errors = validator(config, schema); // TODO: add version 1 to the schema in base repo
```
djmitche: ..and to the docs :)

src/intree.js
owlishDeveloper: Thank you Dustin! This is very helpful. I actually tried taking an initial stub at context, and got tangled :) I'll try hack on the context some time this week.

owlishDeveloper:
Decided to push my work so far just to demonstrate I'm alive - there's nothing to review yet, I don't think.

owlishDeveloper:
https://github.com/taskcluster/taskcluster-github/pull/247

PR #243, opened 2018-05-01
Show Markdown correctly in error comments

djmitche:
Without this patch, the error looks like:

screenshot-2018-5-1 bug 1446966 - try this one simple trick to improve your testing quality by djmitche pull request 99

the issue is that the triple-backticks are put around the Markdown output, causing triple-backticks in the markdown to leave verbatim mode.  And all sorts of madness with emoji.

The taskcluster-lib-api update also avoids the unnecessary backslash in the scope.

### DIFF
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -132,10 +132,10 @@ class Handlers {
     }
     let body = [
       '<details>\n',
-      '<summary>Submitting the task to Taskcluster failed. Details</summary>\n\n',
-      '```js\n',
-      errorBody,
-      '```\n',
+      '<summary>Submitting the task to Taskcluster failed. Details</summary>',
+      '',
+      errorBody, // already in Markdown..
+      '',
       '</details>',
     ].join('\n') ;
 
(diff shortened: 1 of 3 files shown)


src/handlers.js
```diff
@@ -132,10 +132,10 @@ class Handlers {
     }
     let body = [
       '<details>\n',
-      '<summary>Submitting the task to Taskcluster failed. Details</summary>\n\n',
-      '```js\n',
-      errorBody,
-      '```\n',
+      '<summary>Submitting the task to Taskcluster failed. Details</summary>',
+      '',
```
owlishDeveloper, line 136: Will that empty string render as a break between Summary element and the `errorBody`? Maybe a `<br>` in there instead? 

(If we need a break between them, that is)

owlishDeveloper (changes requested):

src/handlers.js
djmitche: It will render as two newlines, which Markdown treats as a paragraph separator.

owlishDeveloper (approved):

PR #244, opened 2018-05-16
Adding a new schema for taskcluster.yml v1

owlishDeveloper:
As a part of the JSON-e-fication project, I think I need to add a new schema to validate new `taskcluster.yml`s v1. After finishing (or preliminary finishing) the documentation, I began implementing. TDD is my technique of choice here, but it seems to me I need the schema to be _up there_ somewhere (I am having trouble reading it from a nearby directory). The tests, I imagine, should validate the YAMLs.

My understanding is, in order to get a schema _up there_, I need to add a yaml file to that effect to the `schemas` dir.

So, the new schema for your consideration!

### DIFF
--- a/schemas/taskcluster-github-config.v1.yml
+++ b/schemas/taskcluster-github-config.v1.yml
@@ -1,12 +1,12 @@
 $schema:            http://json-schema.org/draft-06/schema#
 title:              ".taskcluster.yml format"
 description: |
- Description of a taskclusterrc.yml file v1, which may be used to generate a taskgraph
+ Description of a taskcluster.yml file v1, which may be used to generate a taskgraph
  and tasks.
 type:               object
 properties:
   version:
-    description:    ".taskcluster.yml version"
+    description:    "Version of the format of this file; must be 1"
     enum:           [1]
     type:           integer
   policy:
@@ -25,7 +25,7 @@ properties:
     items:
         title:              "Task definition template"
         description: |
-          Definition of a task that can be scheduled
+          Definition of a task that can be scheduled. Rendered with JSON-e
         type:               object
         additionalProperties: true
 additionalProperties: false
--- a/schemas/taskcluster-github-config.v1.yml
+++ b/schemas/taskcluster-github-config.v1.yml
@@ -1,0 +1,36 @@
+$schema:            http://json-schema.org/draft-07/schema#
+title:              ".taskcluster.yml format"
+description: |
+ Description of a taskclusterrc.yml file v1, which may be used to generate a taskgraph
+ and tasks.
+type:               object
+properties:
+  version:
+    description:    ".taskcluster.yml version"
+    enum:           [1]
+    type:           integer
+  policy:
+    pullRequests:
+      description: |
+          Policy for creating tasks for pull requests.  The effective policy is found in this property
+          in the `.taskcluster.yml` file in the repository's default branch.  See the documentation for
+          detailed definition of the options.
+      type: string
+      enum:
+        - public
+        - collaborators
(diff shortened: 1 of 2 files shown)


schemas/taskcluster-github-config.v1.yml
```diff
@@ -0,0 +1,36 @@
...
+          detailed definition of the options.
+      type: string
+      enum:
+        - public
+        - collaborators
+  tasks:
+    type:           array
+    default:        []
+    items:
+        title:              "Task Definition"
```
imbstack: I think these will be better called something like "task definition template" maybe? They don't need to be task definitions until after this is rendered. Similarly with the `created` and `deadline` fields. Probably don't need to be asserted by this schema? I might be missing something though!

imbstack (approved):
Little nitpick I have but otherwise looks good!

schemas/taskcluster-github-config.v1.yml
owlishDeveloper: I gave it a thunk, and it seems to me we don't have to assert anything. `created` comes easily from context, we can have some default deadline... Why did we assert them in v0?

djmitche: In hooks, we call these a "task template", so we could do the same here.

We also had to remove asserting created/deadline/expires in hooks.  The template can create all three easily enough with `$fromNow`.

schemas/taskcluster-github-config.v1.yml
```diff
@@ -0,0 +1,33 @@
+$schema:            http://json-schema.org/draft-06/schema#
+title:              ".taskcluster.yml format"
+description: |
+ Description of a taskclusterrc.yml file v1, which may be used to generate a taskgraph
```
djmitche: I don't know if this really matters, but that filename is wrong :)

schemas/taskcluster-github-config.v1.yml
```diff
@@ -0,0 +1,33 @@
...
+ and tasks.
+type:               object
+properties:
+  version:
+    description:    ".taskcluster.yml version"
```
djmitche: "Version of the format of this file; must be 1"

schemas/taskcluster-github-config.v1.yml
```diff
@@ -0,0 +1,33 @@
...
+      enum:
+        - public
+        - collaborators
+  tasks:
+    type:           array
+    default:        []
+    items:
+        title:              "Task definition template"
+        description: |
+          Definition of a task that can be scheduled
```
djmitche: It's worth mentioning this is rendered with JSON-e, and linking to the main document.

schemas/taskcluster-github-config.yml
```diff
@@ -1,7 +1,7 @@
 $schema:            http://json-schema.org/draft-06/schema#
 title:              ".taskcluster.yml format"
 description: |
- Description of a taskclusterrc.yml file, which may be used to generate a taskgraph
+ Description of a taskclusterrc.yml file v0, which may be used to generate a taskgraph
```
djmitche: Heh, this filename is wrong too :)

djmitche (approved):
A few more minor nitpicks, but ++

schemas/taskcluster-github-config.v1.yml
owlishDeveloper: I'll mention JSONe, but will add the link in the documentation PR

owlishDeveloper:
This can be merged, right? 🙏

imbstack:
tada!

PR #245, opened 2018-05-21
Bug 1462761 - update to use rootUrl

djmitche:
This also overhauls tc-github to follow [testing best practices](https://docs.taskcluster.net/manual/design/devel/best-practices/testing).

Note that tests won't pass until https://github.com/taskcluster/pulse-publisher/pull/29 and https://github.com/taskcluster/taskcluster-lib-api/pull/99 and https://github.com/taskcluster/taskcluster-client/pull/95 land, are relased, and the corresponding upgrades made here.


test/webhook_test.js
```diff
@@ -1,18 +1,22 @@
...
+
+helper.secrets.mockSuite('webhook', ['taskcluster'], function(mock, skipping) {
+  helper.withEntities(mock, skipping);
+  helper.withFakeGithub(mock, skipping);
+  helper.withServer(mock, skipping);
 
   let github = null;
 
-  setup(async () => {
+  setup(async function() {
```
owlishDeveloper, line 11: why ditch arrow syntax?

owlishDeveloper (approved):

test/webhook_test.js
djmitche: The mocha functions all pass a context that is useful for things like `this.pending = true`, and to get a context you need to use `function` and not an arrow.  Most of the time it doesn't matter, but it's easier to just be consistent about it.

imbstack: @owlishDeveloper it is just a funny mocha-specific thing https://mochajs.org/#arrow-functions

owlishDeveloper: wow, I didn't know this! ok  👍

imbstack (approved):
This is great. Note that you will need to rebase now since I landed @owlishDeveloper changes. 😅

owlishDeveloper:
😊 😇

djmitche:
I've upgraded this to use pulse-publisher 10.0.0 as well.  I'd ask for another look, but there's really not much to see, so unless there's objection I'll probably land as-is, with the existing :heavy_check_mark:

djmitche:
I can't reproduce those CI failures, even in a local `node:8` docker container...

djmitche:
Oh, it's because they're running in parallel and use the same Azure tables, I think.

PR #247, opened 2018-06-14
New and improved JSONe-fication

owlishDeveloper:


src/intree.js
```diff
@@ -103,72 +104,101 @@ module.exports.setup = async function({cfg, schemaset}) {
...
+        throw new Error('Cannot have unicode in branch names!');
+      }
+      if (!branchTest.test(payload.branch || '')) {
+        throw new Error('Cannot have unicode in branch names!');
+      }
+
+      config = jsone(config, {
+        tasks_for: payload.tasks_for,
+        event: payload.body,
+        branch: payload.branch,
```
djmitche: I'd like to not (yet?) have a branch in the context.  Removing this will just mean removing `branch` from some of the `$if` conditions in the fixtures.  If some of those fixtures become redundant, it's OK to drop them -- v1 is simpler than v0, so it makes sense that it requires fewer tests.

djmitche (approved):
Looking good!  (at least) two more things to do:

 * `as_slugid` in the context
 * set `config.scopes` the same way `completeInTreeConfig` does for v0

djmitche (changes requested):
whoops, didn't mean to approve :(

djmitche:
Awesome, just as_slugid left!

djmitche:
(and it looks like one of the intree_test.js tests is failing, but it's not a hard fix)

src/intree.js
```diff
@@ -77,6 +77,37 @@ function completeInTreeConfig(config, payload) {
...
+function createScopes(config, payload) {
+  if (payload.tasks_for === 'github-pull-request') {
+    config.scopes = [
+      `assume:repo:github.com/${ payload.organization }/${ payload.repository }:pull-request`,
+    ];
+  } else if (payload.tasks_for === 'github-push') {
+    let prefix = `assume:repo:github.com/${ payload.organization }/${ payload.repository }:branch:`;
+    config.scopes = [
+      prefix + payload.details['event.base.repo.branch'],
+    ];
```
djmitche: For tags, I think we don't want this scope (just the tag scope)

src/intree.js
```diff
@@ -8,6 +8,15 @@ const jsone = require('json-e');
 
 // Assert that only scope-valid characters are in branches
 const branchTest = /^[\x20-\x7e]*$/;
+let slugids = {};
```
djmitche: This needs to be scoped to the run of the intree function -- putting it here scopes it to the lifetime of the tc-github process, which is too long :)

djmitche (changes requested):

djmitche (approved):
:+1:

imbstack (approved):
Really awesome. I gave this a pretty good read over and see nothing wrong!

PR #249, opened 2018-06-24
Choose validation schema depending on version

owlishDeveloper:
To fix [this]( https://github.com/taskcluster/taskcluster-github/commit/610c4e5db8908ac400eabc5381702fb0f7de2a3a#commitcomment-29424780), I split validation path for the two versions as well. I decided not to touch the `lib-validation` - I'm not sure how it's used elsewhere. In the future, when we ditch the `v0`, this should be an easy refactor to remove the schemas, the choice, and make `handlers` pass the `v1` schema URL into `intree`.

### DIFF
--- a/src/intree.js
+++ b/src/intree.js
@@ -127,6 +127,7 @@ module.exports.setup = async function({cfg, schemaset}) {
     }
   };
   const validate = await schemaset.validator(cfg.taskcluster.rootUrl);
+
   return function({config, payload, schema}) {
     config = yaml.safeLoad(config);
     let slugids = {};
@@ -138,7 +139,14 @@ module.exports.setup = async function({cfg, schemaset}) {
         return slugids[label] = slugid.nice();
       }
     };
-    let errors = validate(config, schema);
+    let version = config.version;
+    
+    let errors;
+    if (version == 0) {
+      errors = validate(config, schema);
+    } else {
+      errors = validate(config, schema.replace(/\.ya?ml$/, '.v1.yaml'));
+    }
     if (errors) {
       throw new Error(errors);
     }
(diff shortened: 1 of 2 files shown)


owlishDeveloper:
I was hoping the build would run on this branch with v1 😿

owlishDeveloper:
oh it's probably because my branch is on my fork, not here (?)

edit: or does `policy` work at all? 🤔 imma check tomorrow

edit2: The code isn't deployed, why am I expecting validation to work 😜
Reverted the commit that replaced the yml

imbstack (approved):
Hah, clever little hack. It might be nicer in the future to check `schema.version` outside this function and pass in different schemas depending on the result. Or we could also pass in a dictionary of schemas and pick one depending on version? I'm going to land this now though!

src/intree.js
```diff
@@ -138,15 +139,21 @@ module.exports.setup = async function({cfg, schemaset}) {
...
       }
     };
-    let errors = validate(config, schema);
+    let version = config.version;
+    
+    let errors;
+    if (version == 0) {
+      errors = validate(config, schema);
+    } else {
+      errors = validate(config, schema.replace(/\.ya?ml$/, '.v1.yaml'));
```
djmitche, line 148: So basically, someone calls this function with "schema", and it decides to use a different schema.

I'm glad this got merged, but this would be a good thing to follow up on in a refactor.

owlishDeveloper: I agree that a refactor is needed here. It seems like to do this properly I would need to modify `lib-validate`? Brian is suggesting some good ideas there

djmitche: Perhaps - or perhaps just re-thinking the "in-tree" abstraction in general?  What is this function's job, and why do we tell it what schema to expect?  Defining that better might make it clearer how to handle this.

PR #251, opened 2018-06-29
V1-v0 split refactor

owlishDeveloper:
I am starting the refactor that would properly split `v1` and `v0` code, one PR at a time. This PR addresses validation. I am switching to Brian's great idea of using dictionaries.

### DIFF
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -314,7 +314,10 @@ async function jobHandler(message) {
       config: repoconf,
       payload: message.payload,
       validator: context.validator,
-      schema: libUrls.schema(this.rootUrl, 'github', 'v1/taskcluster-github-config.yml'),
+      schema: {
+        0: libUrls.schema(this.rootUrl, 'github', 'v1/taskcluster-github-config.yml'),
+        1: libUrls.schema(this.rootUrl, 'github', 'v1/taskcluster-github-config.v1.yml'),
+      },
     });
     if (graphConfig.tasks.length === 0) {
       debug(`intree config for ${organization}/${repository} compiled with zero tasks. Skipping.`);
--- a/src/intree.js
+++ b/src/intree.js
@@ -108,28 +108,28 @@ function createScopes(config, payload) {
 }
 
 /**
- * Returns a function that merges an existing taskcluster github config with
- * a pull request message's payload to generate a full task graph config.
- *  params {
- *    config:             '...', A yaml string
- *    payload:            {},    GitHub WebHook message payload
- *    schema:             url,   Url to the taskcluster config schema
- *  }
- **/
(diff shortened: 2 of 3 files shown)


djmitche (changes requested):
This is looking good!  A few points:

* The middle commit should probably be squashed into the first.
* The `src/handlers.js` change in the third commit should probably be squashed into the first, too.  A quick way to do that is to do a `git rebase -i` and choose to edit the target (first) commit.  Then make a patch containing the bit you want to pull (`git show 03d9960 src/handlers.js > p.patch`) then apply the patch (`patch -p1 < p.patch`) and remove it (`rm p.patch`).

This is heading in the direction of keying everything by the version.  That's shortening the `setup` method nicely.  But in a way you're re-implementing classes and polymorphism:

``javascript
const someMethod = {
  0: (..) => { .. },
  1: (..) => { .. },
};

const someOtherMethod = {
  0: (..) => { .. },
  1: (..) => { .. },
};

let thing = { .. };
someMethod[thing.version](thing);
someOtherMethod[thing.version](thing);
```
works about the same as
```javascript
class Thing {
  static new(version, data) {
    if (version === 0) {
      return new VersionZero(data);
    } else {
      return new VersionOne(data);
    }
  }
}

class VersionZero extends Thing {
  someMethod(..) { .. }
  someOtherMethod(..) { .. }
}
class VersionOne extends Thing {
  someMethod(..) { .. }
  someOtherMethod(..) { .. }
}

thing = Thing.new(..);
thing.someMethod();
thing.someOtherMethod();
```

One benefit of the class arrangement is, you can put the different classes in different files and thus not be distracted by old versions while working on a new feature.

It might make sense to rewrite the last commit in this series to use classes now.  Or, it might make sense to just use this PR to separate out the version-specific bits, and leave use of classes to another PR.

"Request changes" just for the squashing -- I'm happy with the overall diff in this PR.

owlishDeveloper:
Classes sound good!

owlishDeveloper:
Squashes 🎃 performed! Please take a look if all the changes are in correct commits 🙏

I would like to implement the classes idea separately from this PR.

Thank you for the review, Dustin!

owlishDeveloper:
(Are you sure the classes won't make the refactor harder once we get rid of version 0?)

djmitche (approved):
Looks good to me.

If I've learned one thing, it's that I'm not sure of anything.  But I think having a class for v0 and a class for v1 means that, when we decide to get rid of v0, we can just delete the class and call it finished.

owlishDeveloper:
Sounds good! 😊

imbstack (approved):
This looks great. I will land this today and we can deploy monday!

PR #252, opened 2018-07-05
Scope mess error message should contain some hint how to fix it

edunham:
I asked Dustin and he explained how to use https://tools.taskcluster.net/auth/roles to debug this kind of thing. 

I don't care much at all about the wording, format, placement, etc of the message, just that a small hint is somewhere in the error.

### DIFF
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -125,6 +125,7 @@ class Handlers {
       '',
       errorBody, // already in Markdown..
       '',
+      'Try using the role editor at tools.taskcluster.net/auth/roles to correct these problems.',
       '</details>',
     ].join('\n') ;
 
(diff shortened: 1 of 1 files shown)


src/handlers.js
```diff
@@ -125,6 +125,7 @@ class Handlers {
       '',
       errorBody, // already in Markdown..
       '',
+      'Try using the role editor at tools.taskcluster.net/auth/roles to correct these problems.',
```
djmitche, line 128: This kind of suggestion would make sense if conditionalized on the error code being `InsufficientScopesError` - other errors wouldn't be fixed this way.

The error should not refer to a specific hostname, though, since in another deployment the hostname would be different.  There is a library ([taskcluster-lib-urls](https://docs.taskcluster.net/docs/reference/libraries/taskcluster-lib-urls), imported as libUrls) that can generate UI links specific to the deployment.

It would also be good to suggest *which* roles to edit. Since there may be several matching the role with `*`, it's probably best to just describe the full roleId (from `graphConfig.scopes`) and suggest that there may be several roles defined matching it.

djmitche (changes requested):

petemoore:
@edunham Are you still interested in working on this?

owlishDeveloper:
This probably can be closed?

djmitche:
I think so..

PR #253, opened 2018-07-09
Start using v1 .taskcluster.yml

imbstack:
This just switches out our own use of v0 for v1! Slight tweak needed to get it to work.

I have deployed `master` back to production so this won't work for now.

### DIFF
--- a/src/intree.js
+++ b/src/intree.js
@@ -166,26 +166,26 @@ compileTasks = {
       if (!task.task.extra || !task.task.extra.github) {
         return false;
       }
-  
+
       let event = payload.details['event.type'];
       let events = task.task.extra.github.events;
       let branch = payload.details['event.base.repo.branch'];
       let includeBranches = task.task.extra.github.branches;
       let excludeBranches = task.task.extra.github.excludeBranches;
-  
+
       if (includeBranches && excludeBranches) {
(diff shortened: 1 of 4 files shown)


owlishDeveloper:
I'm all wrong about the payload shape, it seems! 😄 I'll take a look, thank you Brian!

src/intree.js
```diff
@@ -211,11 +211,15 @@ compileTasks = {
...
+      const groupId = slugid.nice();
       config.tasks = config.tasks.map((task) => {
         if (!task.taskId) { throw Error('The taskId is absent.'); }
         return {
           taskId: task.taskId,
-          task: _.extend(task, {schedulerId: cfg.taskcluster.schedulerId}),
+          task: _.omit(_.extend(task, {
+            taskGroupId: task.taskGroupId || groupId,
+            schedulerId: cfg.taskcluster.schedulerId,
+          }), 'taskId'),
```
djmitche, line 222: This bit looks good!

djmitche (approved):

src/intree.js
owlishDeveloper: is it so that we don't have an object with `null` properties?

EDIT: got it! no `taskId`. It's already up there!

owlishDeveloper (approved):

PR #254, opened 2018-07-14
Classy Refactor

owlishDeveloper:
Follow up on `v1` vs `v0` refactor - I pulled all the `yml`-related logic in a separate module, and grouped the functions by version using classes. Naming things is a hard problem, so I'm opened to suggestions there especially 😜

Also, I'm not sure about file organization - maybe `v0` stuff should be in a separate file?


src/tc-yaml.js
```diff
@@ -0,0 +1,234 @@
...
+const tc = require('taskcluster-client');
+
+// Assert that only scope-valid characters are in branches
+const branchTest = /^[\x20-\x7e]*$/;
+
+/**
+ * Get scopes and attach them to the task.
+ * v1 function
+ */
+function createScopes(config, payload) {
```
djmitche: This function and the next probably belong in the VersionOne and VersionZero classes.

src/tc-yaml.js
```diff
@@ -0,0 +1,234 @@
...
+          GITHUB_HEAD_USER_EMAIL: payload.details['event.head.user.email'],
+        }
+      );
+    }
+    return task;
+  });
+  return config;
+};
+
+class Tcyaml {
```
djmitche: Note that this class is required as `TcYaml` in `intree.js`.  It'd be good to be consistent in the capitalization.

src/intree.js
```diff
@@ -257,13 +34,14 @@ module.exports.setup = async function({cfg, schemaset}) {
     // Perform parameter substitutions. This happens after verification
     // because templating may change with schema version, and parameter
     // functions are used as default values for some fields.
-    config = substituteParameters[version](config, cfg, payload);
+    const yml = TcYaml.instantiate(version);
```
djmitche: Let's call this `tcyaml`, to better align with the class name.

djmitche (approved):
Minor suggestions, but this looks like a good refactor!

imbstack (approved):
Makes sense to me after dustin's changes!

djmitche:
Looks good -- @imbstack do you want to deploy today?

PR #255, opened 2018-07-16
Open up validation of json-e templates

imbstack:
I think this should allow sufficient leeway for people to use json-e effectively.


djmitche (approved):
Yeah, that's the idea :)

PR #260, opened 2018-07-19
Bug 1476452 - apply defaults for taskGroupId, taskId

djmitche:


djmitche:
This adds some tests for `tc-yaml.js`, too.  At least for the part I changed :)

owlishDeveloper (approved):
Looks very good to me!

owlishDeveloper:
(Not sure if related... What is happening to the recent builds? Seems to be a problem with task group id?)

djmitche:
I think it is related :)

djmitche:
released

PR #262, opened 2018-07-20
Clarify event payload

jugglinmike:
In version 3 of the GitHub API, the payload of the "Events API" differs
from that of the "Webhook" API. This is discussed in the description of
"PushEvent" [1]:

> Note: The webhook payload example following the table differs
> significantly from the Events API payload described in the table.
> Among other differences, the webhook payload includes both sender and
> pusher objects. Sender and pusher are the same user who initiated the
> push event, but the sender object contains more detail.

Clarify this in the description of the rendering context.

https://developer.github.com/v3/activity/events/types/#pushevent

### DIFF
--- a/docs/taskcluster-yml-v1.md
+++ b/docs/taskcluster-yml-v1.md
@@ -50,7 +50,7 @@ The `tasks` property in the YAML file is rendered using [JSON-e](https://github.
 * `tasks_for` - defines the type of event, one of `github-push`,
   `github-pull-request` or `github-release`.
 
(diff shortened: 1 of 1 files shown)


owlishDeveloper (approved):

djmitche:
This is a good fix, but I admit I find that github doc baffling.  Basically it says "The docs and the example here don't agree.  So it's up to you to guess which one you will get."

djmitche:
I filed a support request with github :)

PR #264, opened 2018-07-23
Don't log the github token to the console

djmitche:
We don't want secrets in our logs, as they tend to get copy-pasted.


djmitche:
Not sure why I never asked for review..

imbstack (approved):

PR #266, opened 2018-07-26
Support the 'policy' property in v1

djmitche:
Looks like this got forgotten in the v1 setup :)

It explains why PRs are getting errors, like https://github.com/taskcluster/taskcluster-cli/pull/205


imbstack (approved):
Other than the failing tests I think this makes sense

djmitche:
they're failing on master :(

djmitche:
oh, that was my fault from #265

PR #267, opened 2018-08-15
Documentation and context updates

owlishDeveloper:
As per discussion [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1476856#c5), I am updating v1 documentation so that it'd be easier to look up github webbhook payload things when moving a repo from v0 to v1 (or even creating a v1 yaml from scratch).

We also agreed that `created` doesn't need to be specified each time, so I'm adding it to the context.


djmitche (approved):
Awesome :)

imbstack (approved):

Issue #271, opened 2018-09-12
Docs: .taskcluster.yml v1 example is missing "created" field


JohanLorenzo:
The example provided in https://docs.taskcluster.net/docs/reference/integrations/taskcluster-github/docs/taskcluster-yml-v1 doesn't list the required field `created`. See https://github.com/mozilla-mobile/fenix/pull/64#commitcomment-30478953

djmitche:
Hm, #267 did that intentionally, but it looks like it added `created` to the context, rather than adding it to the task.  @owlishDeveloper can probably take a look :)

owlishDeveloper:
L👀king :)

owlishDeveloper:
What I'm not sure about, however, is why in that PR https://github.com/mozilla-mobile/fenix/pull/64 the `v1` was evaluated against `v0` schema 🤔

djmitche:
Is it possible that we look for the version in the master branch (where we look for the PR policy) instead of in the file itself?

owlishDeveloper:
💯

owlishDeveloper:
Solved

PR #272, opened 2018-09-13
Fixes `created` bug and adds `$match` to the documentation

owlishDeveloper:
https://github.com/taskcluster/taskcluster-github/issues/271


docs/taskcluster-yml-v1.md
```diff
@@ -110,6 +110,16 @@ tasks:
...
+OR
+```yaml
+version: 1
+tasks:
+  $match:
+    'tasks_for == "github-push"':
+      ...
+      ...
+```
+
```
djmitche, line 122: :100:

test/tc-yaml_test.js
```diff
@@ -26,6 +26,7 @@ suite('tc-yaml_test.js', function() {
       assume(config.tasks).to.deeply.equal([{
         taskId: config.tasks[0].taskId,
         task: {
+          created: new Date().toJSON(),
```
djmitche: This doesn't pass :(

> Unknown assertation failure occured, assumed `[ { task: { created: '2018-09-13T15:24:27.542Z', schedulerId: 'test-sched', taskGroupId: 'fxVScA_cRdejLPnJXtKFrQ' }, taskId: 'fxVScA_cRdejLPnJXtKFrQ' } ]` to deeply equal `[ { task: { created: '2018-09-13T15:24:27.544Z', schedulerId: 'test-sched', taskGroupId: 'fxVScA_cRdejLPnJXtKFrQ' }, taskId: 'fxVScA_cRdejLPnJXtKFrQ' } ]`

Note that the milliseconds on the timestamps are different.

Usually the best fix for this sort of problem is to pass in a single value for `now` that the test can specify and then compare.  Something like

```
const now = new Date();
tcyaml.compileTasks(config, cfg, {}, now);
assume(config.tasks).to.deeply.equal(... created: now.toJSON() ...);
```

I don't recall if we create a new TcYaml instance for every new Github event.  If we do, then it might make more sense to set `now` in the VersionOne constructor.  Then there's no need to pass `now` to `compileTasks`, and the tests can just access the `tcyaml.now` value for checks:

```
assume(config.tasks).to.deeply.equal(... created: tcyaml.now.toJSON() ...);
```

djmitche (changes requested):

test/tc-yaml_test.js
owlishDeveloper: Thank you Dustin!

src/tc-yaml.js
```diff
@@ -215,7 +215,6 @@ class VersionOne extends TcYaml {
         tasks_for: payload.tasks_for,
         event: payload.body,
         as_slugid,
-        created: new Date().toJSON(),
```
djmitche, line 218: Happily, this was never documented, so I think removing it is OK.

djmitche (approved):
This looks good.  Ship it!

PR #273, opened 2018-09-13
Enforce taskGroupId to match taskId when 1 task exists

JohanLorenzo:
Follows #260 up

[Documentation reads](https://docs.taskcluster.net/docs/reference/integrations/taskcluster-github/docs/taskcluster-yml-v1#result): 

>  If the JSON-e rendering produces only one task, then the default taskGroupId and taskId have the same value

There is a case in the current implementation where this statement is not valid. If you set a `taskId` on the only task defined in `.taskcluster.yml`, then the `taskGroupId` doesn't match the `taskId`.

This patch fixes it.

r? @djmitche


djmitche (approved):
Nice!

PR #274, opened 2018-09-28
[WIP] Add Checks to tc-gh

owlishDeveloper:


owlishDeveloper:
Actually, looking at this - I will be upgrading `github` dependency in a separate PR which I will open tomorrow.

src/api.js
```diff
@@ -187,7 +186,7 @@ builder.declare({
   stability:  'experimental',
   description: [
     'Capture a GitHub event and publish it via pulse, if it\'s a push,',
-    'release or pull request.',
+    'release or pull request. We also publish ping and checks events',
```
djmitche: We do not seem to actually publish ping events..

schemas/v1/github-checkrun-message.yml
```diff
@@ -0,0 +1,27 @@
+$schema:  http://json-schema.org/draft-06/schema#
+title:        "GitHub Check Run Message"
+description: |
+  Message reporting that a GitHub check run has occurred
+type:         object
+
+properties:
```
djmitche: Shouldn't this have something about the check that ran, too?

src/api.js
```diff
@@ -286,16 +295,17 @@ builder.declare({
...
   // Not all webhook payloads include an e-mail for the user who triggered an event
-  let headUser = msg.details['event.head.user.id'].toString();
+  let headUser = msg.body.sender.id.toString();
   let userDetails = (await instGithub.users.getById({id: headUser})).data;
-  msg.details['event.head.user.email'] = this.ajv.validate({type: 'string', format: 'email'}, userDetails.email) ?
-    userDetails.email :
-    msg.details['event.head.user.login'].replace(/\[bot\]$/, '') + '@users.noreply.github.com';
+  msg.details['event.head.user.email'] = this.ajv.validate({type: 'string', format: 'email'}, userDetails.email)
+    ? userDetails.email
+    : msg.body.sender.login.replace(/\[bot\]$/, '') + '@users.noreply.github.com';
```
djmitche, line 302: This bit looks unrelated to checks?

src/api.js
```diff
@@ -505,7 +515,7 @@ builder.declare({
     } catch (e) {
       // 403 from Github indicates this integration doesn't have permission to post this status,
       // so return that on to the user
-      if (e.code == 403) {
+      if (e.code === 403) {
```
djmitche: it'd be good to land all of these fixes (===, extra semicolons, etc.) in a single PR.  I suspect your editor is auto-fixing these?  They're distracting and have the potential to cause merge conflicts when mixed into other, more substantive PRs.

djmitche (commented):
It looks like the overall idea here is just to get check run webhooks "wired up" to sending pulse messages -- but not actually do anything with those pulse messages?  That seems reasonable, but what is the bigger picture that this fits into?

schemas/v1/github-checkrun-message.yml
owlishDeveloper: yeah, this part is not finished yet

src/api.js
owlishDeveloper: No, it's me who auto-fixes these. Yeah, I can totally land these in a separate PR (I'll throw couple of other similar things in that PR then, actually)

djmitche: We could certainly enable a few more eslint checks across the board (for all of our JS repos) to prevent re-adding errors like these.  But that's a whole different project :)

src/api.js
owlishDeveloper: Yeah, you're right - I'll gather these to a separate PR.

src/api.js
owlishDeveloper: Adding couple of rules to `eslint-config-taskcluster` seems like an easy thing to do? I can do that

src/api.js
owlishDeveloper: so, since most of the syntax changes in this PR are left for a "good first bug", and there's no eslint rule that enforces ternary operator formatting, I'll be leaving this particular change in this PR

owlishDeveloper:
phew, cleaned up this PR 😅 Now all the irrelevant things are gone 👍

owlishDeveloper:
The bigger picture here is that after we got the `check_suite` message out of the pulse, we will be creating check runs for each of the tasks. Those check runs will be added to the suite by github, and then it should Just Work... or something along these lines. I would try to avoid any concurrency though. I'm going to see if I can re-animate tc-cat and play with it a bit tomorrow to examine these webhooks a bit closer

owlishDeveloper:
I'm going to close this PR for a while, so that it doesn't bother anybody.

PR #275, opened 2018-10-02
[WIP] Upgrade github dependency to allow for Checks API

owlishDeveloper:


djmitche (approved):
I assume the interface remains the same, just a name-change?

It looks like `package-lock.json` got added here -- npm does that.  We prefer yarn (`yarn.lock`).  So please remove `package-lock.json` before landing.

owlishDeveloper:
That's what I assume as well. I am looking through their source code history at the moment

Oh, I thought it would be nice to have both of them, in case any of us or outside collaborators runs `npm install` instead of `yarn`?

djmitche:
I think keeping them in sync would be problematic.  We definitely recommend `yarn` all the time at the moment.

owlishDeveloper:
Yeah, I was just reading about this on the internets - looks like you're right! I'll get rid of it.

I checked all the interfaces that we use - they seem to be the same in the new version.

owlishDeveloper:
(pay no attention to that commit, I'll sort it out in a moment)

djmitche (approved):

PR #278, opened 2018-10-10
[WIP] Bug 1459645 - use github checks

owlishDeveloper:
This is the first iteration of checks. With these changes, we will be displaying statuses via checks api. The next step would be to make the retrigger task button functional, and then we can even add some custom actions for our users (if we want to)

100% review please!

### DIFF
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -116,7 +116,7 @@ class Handlers {
   async createExceptionComment({instGithub, organization, repository, sha, error, pullNumber}) {
     let errorBody = error.body && error.body.error || error.message;
     // Let's prettify any objects
-    if (typeof errorBody == 'object') {
+    if (typeof errorBody === 'object') {
       errorBody = JSON.stringify(errorBody, null, 4);
     }
     let body = [
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -213,7 +213,7 @@ async function groupStatusHandler(message) {
 
   await build.modify(b => {
     if (b.state !== 'failure') {
-      b.taskState = groupState;
+      b.state = groupState;
       b.updated = new Date();
     }
   });
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -478,9 +478,8 @@ async function jobHandler(message) {
       checkSuiteId: checkRun.data.check_suite.id.toString(),
       checkRunId: checkRun.data.id.toString(),
     }).catch(async (err) => {
-      if (err.code !== 'EntityAlreadyExists') {
-        throw err;
-      }
+      await this.createExceptionComment({instGithub, organization, repository, sha, error: err});
+      throw err;
     });
   }));
 
@@ -498,6 +497,7 @@ async function jobHandler(message) {
     eventId: message.payload.eventId,
   }).catch(async (err) => {
     if (err.code !== 'EntityAlreadyExists') {
+      await this.createExceptionComment({instGithub, organization, repository, sha, error: err});
       throw err;
     }
     let build = await this.Builds.load({
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -5,6 +5,7 @@ const yaml = require('js-yaml');
 const assert = require('assert');
 const _ = require('lodash');
 const prAllowed = require('./pr-allowed');
+const {consume} = require('taskcluster-lib-pulse');
 
 const INSPECTOR_URL = 'https://tools.taskcluster.net/task-group-inspector/#/';
 
@@ -33,7 +34,7 @@ const debug = Debug(debugPrefix);
  * Create handlers
  */
 class Handlers {
-  constructor({rootUrl, credentials, monitor, reference, jobQueueName, statusQueueName, intree, context}) {
+  constructor({rootUrl, credentials, monitor, reference, jobQueueName, statusQueueName, intree, context, pulseClient}) {
     debug('Constructing handlers...');
     assert(monitor, 'monitor is required for statistics');
     assert(reference, 'reference must be provided');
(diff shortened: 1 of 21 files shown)


src/handlers.js
```diff
@@ -167,38 +195,53 @@ async function statusHandler(message) {
...
-
-  if (message.exchange.endsWith('task-exception') || message.exchange.endsWith('task-failed')) {
-    state = 'failure';
-  }
+  } while (params.continuationToken);
 
   await build.modify(b => {
     if (b.state !== 'failure') {
-      b.state = state;
+      b.taskState = groupState;
```
djmitche: I don't see this field renamed in `src/data.js`.

src/handlers.js
```diff
@@ -380,30 +424,54 @@ async function jobHandler(message) {
...
-      repo: repository,
-      sha,
-      state: groupState,
-      target_url: INSPECTOR_URL + taskGroupId,
-      description,
-      context: statusContext,
-    });
+
+    await Promise.all(graphConfig.tasks.map(async (task, i) => {
+      const checkRun = await instGithub.checks.create(Object.assign( // TODO: need spread syntax
```
djmitche: ++ to spread syntax :)

src/handlers.js
```diff
@@ -153,12 +180,13 @@ class Handlers {
...
- * GitHub Statuses: https://developer.github.com/v3/repos/statuses/
- **/
-async function statusHandler(message) {
-  let taskGroupId = message.payload.taskGroupId || message.payload.status.taskGroupId;
+ * Modify the build in the Azure table when the taskgroup is resolved
+ *
+ * @param exchange message from Queue service
+ * @returns {Promise<void>}
+ */
+async function groupStatusHandler(message) {
```
djmitche, line 198: Is it even necessary to listen to task statuses anymore?  I suspect we could instead poll the github API for the checkSuite for the given repo / sha, and not even need the Builds table anymore.

I think keeping one kind of status in the Builds table and one kind in the checkSuite's will cause confusion, where the last commit has a green checkmark (checkSuite OK) but the SVG icon says the build failed (taskgroup is not 100% success).

djmitche (commented):
So, this would replace the status API with a similar effect using the checks API, except that we report per-task status to github via checkRun instead of per-taskgroup status.  Github then summarizes all of those checkRun's into a single status for the push or PR.  From irc conversation, it sounds like this would be equivalent to the current behavior, even for users who are consuming PR status through an API (like homu).  But it would have a bit more data available in the "checks" tab when `.taskcluster.yml` generates more than one task.

src/handlers.js
owlishDeveloper: I opened a PR for that: https://github.com/taskcluster/eslint-config-taskcluster/pull/7

src/handlers.js
owlishDeveloper: you mean group statuses? I wasn't sure because I am best familiar with tc-gh, and not as much with other services, so I don't know how Builds table fits into the overall picture

`CheckRuns` table doesn't keep statuses - it keeps check run IDs, so that we could update the statuses later

owlishDeveloper: Or are you talking about the table you proposed in the channel? Since we have all the check run ids, we don't need to store the statuses - we can ask github

src/handlers.js
owlishDeveloper: good catch! 🐟

src/handlers.js
```diff
@@ -8,6 +8,24 @@ const prAllowed = require('./pr-allowed');
 
 const INSPECTOR_URL = 'https://tools.taskcluster.net/task-group-inspector/#/';
 
+const TITLES = { // maps github checkruns statuses and conclusions to titles to be displayed
+  success: 'Success',
+  failure: 'Failure',
+  neutral: 'It is neither good nor bad',
```
imbstack, line 13: 😆

src/handlers.js
imbstack: I think maybe we can even get rid of the builds table and anything to do with task group status! As you two were discussing in irc, just having a mapping of task -> check makes sense. The `taskgroupresolved` stuff has always been kinda flaky/confusing anyway.

We can probably avoid having a flag by proactively finding which orgs have tc installed and asking them to update our permissions way before we actually make the switch to the checks api. Ideally we don't actually require any changes to the tasks themselves by default?

This would also simplify the handler code drastically!

imbstack (approved):
This is looking great. Let's talk tomorrow on what our plan to roll forward to this would be.

src/handlers.js
```diff
@@ -380,30 +424,54 @@ async function jobHandler(message) {
...
-    });
+
+    await Promise.all(graphConfig.tasks.map(async (task, i) => {
+      const checkRun = await instGithub.checks.create(Object.assign( // TODO: need spread syntax
+        {
+          owner: organization,
+          repo: repository,
+          name: `Task ${i}: ${this.context.cfg.app.statusContext} (${eventType.split('.')[0]})`,
+          head_sha: sha,
+          output: { // TODO: maybe a more helpful output?
```
imbstack: We can probably stick logs from our tasks into this? At least that's my understanding of the api.

src/handlers.js
djmitche: Yes, sorry, I meant "group statuses" in the first comment in this thread, sorry for the confusion.

src/handlers.js
djmitche: I wonder if logs would be too large for github?  I was vaguely thinking that as a follow-up we could define a specific artifact that would be reflected into the status, including output and annotations.  Then users could choose what output to include.

owlishDeveloper:
we still need the Builds table, it seems

owlishDeveloper:
This work continues in #309

PR #280, opened 2018-10-10
Use new eslint. Fix all the syntax

owlishDeveloper:
I know this was planned as a good-first-bug, but I need this done for my checks PR, to use the spread syntax


owlishDeveloper:
I know this was planned as a good-first-bug, but I need this done for my checks PR, to use the spread syntax

djmitche (approved):
That's fine - when a GFB becomes important, just do it :)

PR #282, opened 2018-10-16
Switch to tc pulse

owlishDeveloper:
I am a bit concerned about multiple `call(this, message)` around `callHandler`... There must be a way of simplifying that?


src/main.js
```diff
@@ -77,15 +75,32 @@ const load = loader({
...
 
+  pulseClient: {
+    requires: ['cfg', 'monitor'],
+    setup: ({cfg, monitor}) => {
+      const {namespace, expiresAfter, contact} = cfg.pulse;
+      const {rootUrl, credentials} = cfg.taskcluster;
+      return new Client({
+        namespace,
+        monitor,
+        credentials: claimedCredentials({
```
djmitche: In a late-breaking change, let's use `pulseCredentials` for this instead of claimedCredentials.  That will also go back to using username/password/hostname/vhost.

src/handlers.js
```diff
@@ -85,13 +86,22 @@ class Handlers {
...
-
-    // If this is awaited, it should return [undefined, undefined]
-    await Promise.all([this.jobListener.resume(), this.statusListener.resume()]);
+    this.jobPq = await consume(
+      {
+        client: this.pulseClient,
+        bindings: jobBindings,
+        queueName: this.jobQueueName,
+      },
+      message => this.monitor.timedHandler('joblistener', callHandler('job', jobHandler).call(this, message))
```
djmitche: I think the confusion here is that [`timedHandler`](https://github.com/taskcluster/taskcluster-lib-monitor/blob/7850f27d353a3e049bb15a83343fb32efdc3b18f/src/utils.js#L59-L77) takes a callable object and then calls it, measuring the duration between start and finish of the call.  The `callHandler` function returns a function, written in a way only a functional-programming-languages geek (jonas) could love (`(name, handler) => message => { .. }`).  Basically calling it once with `(name, handler)` returns the sub-function `message => { .. }`.  Which is exactly what we want to pass to `timedHandler`.  Adding an additional `.call(this, message)` calls that second function right away and returns the result, which is then passed to timedHandler.

There's no need to worry about binding context (`this`) here -- `callHandler` is an arrow function so it doesn't interact with context values, and it captures `this` from the surrounding function.

So basically, this should still work with `.call(this, message)` omitted, and with that modification it will actually get timings for handlers.

config.yml
```diff
@@ -41,10 +41,9 @@ defaults:
     trustProxy: false
 
   pulse:
-    username: !env PULSE_USERNAME
-    password: !env PULSE_PASSWORD
-    hostname: !env PULSE_HOSTNAME
-    vhost: !env PULSE_VHOST
+    namespace: !env PULSE_NAMESPACE
+    expiresAfter: !env PULSE_EXPIRES_AFTER
+    contact: !env PULSE_CONTACT
```
djmitche: Related to the late-breaking change, we'll go back to using PULSE_USERNAME, etc. so you can revert this hunk.

djmitche (changes requested):
This looks great other than the two details below, one of which was just decided last night.

src/handlers.js
owlishDeveloper: Yeah, this all made sense to me here; what I can't understand is - why do the tests pass for tc-notify and tc-index? I actually tried passing in a function when I was preparing the PR for tc-index - the tests started failing. When replaced function with a call of a function - the tests worked. I briefly glanced at monitor's source code - it expects a function 🤔 I am lost! (jonas is not the only functional-programming-loving geek around here 😛) 

I think you do need to bind to `this`. For me, when not bound, I was getting a global object in there instead of `this`. The tests were failing to refer to `this.context`, so I `console.log`ed `this` - it was the global object. I'm not sure why the binding doesn't work here - maybe there's non-arrow function there in the chain somewhere. I'll check.

Re: `.call(this, message)` - it doesn't work without it :( I tried all sorts of things. When I passed in `callHandler('job', jobHandler)` - I was getting incorrect `this`. When I was passing in `jobHandler` - incorrect `this` as well, when I tried `jobHandler.bind(this)` I was getting timeouts.

I am going to look at this more closely today. The fact that I don't understand this code concerns me

djmitche: Hm, you're right -- timedHandler should fail if you give it a value instead of a function.  So clearly I've missed something!

djmitche (approved):
This is looking great.  If the handlers are working, and you want to defer the question of why `call` is required to another bug, that's totally cool by me.

djmitche (approved):
I continue to approve of this PR :)

imbstack (approved):

PR #287, opened 2018-10-19
This foxes the queueName assertion error

owlishDeveloper:
(I like the typo in the title, decided to leave it 🦊 )


src/main.js
```diff
@@ -187,8 +187,8 @@ const load = loader({
       monitor: monitor.prefix('handlers'),
       intree,
       reference,
-      jobQueueName: cfg.app.jobQueueName,
-      statusQueueName: cfg.app.statusQueueName,
```
djmitche, line 191: So I guess like this was the actual culprit, then? The queue names were being left at `undefined` which `consume` doesn't like.

djmitche (approved):

src/main.js
owlishDeveloper: Exactly!

PR #288, opened 2018-10-19
Set trustProxy and forceSSL based on env vars

djmitche:
The settings are different in Heroku and in GKE, so they need to be
given by env vars.


imbstack (approved):

djmitche:
(I checked these are set in Heroku)

PR #289, opened 2018-10-22
Bug 1501005 - shut down monitoring when finished syncing

djmitche:
This fixes the hangs, which occur because tc-lib-monitor is running on an interval (`setInterval`) and node is sticking around to service that.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -166,8 +166,8 @@ const load = loader({
   },
 
   syncInstallations: {
-    requires: ['github', 'OwnersDirectory'],
-    setup: async ({github, OwnersDirectory}) => {
+    requires: ['github', 'OwnersDirectory', 'monitor'],
+    setup: async ({github, OwnersDirectory, monitor}) => {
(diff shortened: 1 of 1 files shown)


owlishDeveloper (approved):

src/main.js
```diff
@@ -176,6 +176,8 @@ const load = loader({
           owner: inst.account.login,
         }, true);
       });
+      monitor.stopResourceMonitoring();
```
owlishDeveloper, line 179: oh, this probably needs the `monitor` to be passed in

owlishDeveloper (changes requested):

src/main.js
djmitche: Ack, good catch.  I made this change on a running pod to test it and then failed to copy out all of my changes.

djmitche:
ok, checks pass now :)

owlishDeveloper (approved):

PR #290, opened 2018-10-26
Small fixes

owlishDeveloper:
* Transition from deprecated form of github API to modern
* Something uncovered by me trying to get tc-🐈 work - we have inspector URL hard-coded, so the links to the tasks will be broken everywhere except main cluster.

This is currently being tested by me and tc-🐈

### DIFF
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -7,8 +7,6 @@ const _ = require('lodash');
 const prAllowed = require('./pr-allowed');
 const {consume} = require('taskcluster-lib-pulse');
 
-const INSPECTOR_URL = 'https://tools.taskcluster.net/task-group-inspector/#/';
-
 const debugPrefix = 'taskcluster-github:handlers';
 const debug = Debug(debugPrefix);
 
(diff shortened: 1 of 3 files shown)


djmitche:
> so the links to the tasks will be broken everywhere except main cluster.

:heart: we have a lot of this sort of thing to clean up, but there's no better time to start!

src/handlers.js
```diff
@@ -219,13 +217,14 @@ async function statusHandler(message) {
   }
 
   debug(`Attempting to update status for ${build.organization}/${build.repository}@${build.sha} (${state})`);
+  const target_url = `${this.context.cfg.taskcluster.rootUrl}/task-group-inspector/#/${taskGroupId}`;
```
djmitche, line 220: You can use the `ui` function of `taskcluster-lib-urls` to generate this (it does about the same thing, but more cleanly).

owlishDeveloper: This seems to be simpler to change and debug tho... If there ever be something wrong with this url - you can immediately see where it's coming from

imbstack: Yeah, we're just trying to standardize on the library though so that we can easily make changes in the future! Otherwise this is more readable.

imbstack (approved):

owlishDeveloper:
I'll leave the template literal in for now (I really favor readability...) We can always replace it with a more clever machinery when we need it!

djmitche:
I think ``libUrls.ui(rootUrl, `/task-group-inspector/${..}`)`` is pretty readable..

djmitche:
Worth noting, too, that that's an old URL that redirects to somewhere else now.  I think we want `/tasks/<taskId>`.

owlishDeveloper:
It's readable, but if you need to trace the origins of the url you'll have to follow a whole chain of functions and classes. This way you don't to command-click anywhere, keeping all the previous things in memory - it's right here, all in one place in its entirety. When debugging, I find I prefer things to be that way. I'm a very primitive creature I'm afraid 😩😣

Yeah, it is an old url, but the redirection works correctly. This old URL will go away when I merge checks in

owlishDeveloper:
For example, even using that method: I was, like: what about the initial slash? I assume the method is smart enough to make sure it doesn't matter whether your path starts with a slash or not, but I felt I wanted to make sure, just in case. I realize I'm being obstinate in almost nasty way, but I do prefer to keep things simpler until a more complex solution is really needed

which reminds me I wanted to make another PR

PR #292, opened 2018-10-31
Use `libUrls.ui` method

owlishDeveloper:
So I had thought every service had their own `rootUrl` in the config, so I thought `tools.taskcluster.net` will appear automatically. After all the argument around the `ui` function and gaining understanding what it was for, I realized that the integration will be broken without the `ui` method. I tried deploying the last commit - indeed, broken it was. So I'm switching to that method here!


djmitche (approved):
Hah, yes, the old version wouldn't work right for `https://taskcluster.net`, would it?

owlishDeveloper:
No it wouldn't 😛

PR #294, opened 2018-10-31
Add CI to the context

owlishDeveloper:
🤔

Will be deploying this to my cluster too look at it

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -6,7 +6,7 @@ defaults:
     buildsTableName: 'TaskclusterGithubBuilds'
     ownersDirectoryTableName: 'TaskclusterIntegrationOwners'
     publishMetaData: !env:bool PUBLISH_METADATA
-    statusContext: 'Taskcluster'
+    statusContext: 'taskcluster ci' # "ci" is for third party tools: https://docs.codecov.io/docs/detecting-ci-services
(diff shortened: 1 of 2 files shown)


config.yml
```diff
@@ -6,7 +6,7 @@ defaults:
     buildsTableName: 'TaskclusterGithubBuilds'
     ownersDirectoryTableName: 'TaskclusterIntegrationOwners'
     publishMetaData: !env:bool PUBLISH_METADATA
-    statusContext: 'Taskcluster'
+    statusContext: 'Taskcluster CI' # some tools require an indication that this is a CI service
```
imbstack: My only worry is that maybe they don't downcase this when they read it and will be expecting `ci`? Otherwise, let's ship it and we can ask sebastien to test.

imbstack (approved):

config.yml
djmitche: Let's link to https://docs.codecov.io/docs/detecting-ci-services specifically in the comment.

Note that we could use `continuous-integration/taskcluster` to follow the pattern Travis-CI sets.

owlishDeveloper: link is a good point, I'll add it.

yeah, it crossed my mind as well, the case thing. ok

djmitche (approved):

PR #298, opened 2018-11-06
Add debug messages to jobHandler

owlishDeveloper:
[Re: this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1499576)

I wasn't able to find anything much in the logs as much of our debug messages don't contain shas (which I was given). Besides, something I had noticed even earlier: many of the debug messages are unidentifiable too, as tc-gh can run concurrently.


imbstack (approved):
Awesome. This will be extra good once we move to structured logging stuff!

PR #299, opened 2018-11-07
Checks phase 0 - changing initial status creation flow

owlishDeveloper:
I will be deploying this to my cluster for testing after the initial review pass


src/handlers.js
```diff
@@ -56,18 +71,24 @@ class Handlers {
       githubEvents.release(),
     ];
 
+    let schedulerId = this.context.cfg.taskcluster.schedulerId;
```
imbstack: May as well make these `const` while we're at it.

imbstack: May as well make these `const` while we're at it.

imbstack (approved):
If I've read this correctly it should do exactly what we do now but is laying the ground work for next steps?

owlishDeveloper:
> If I've read this correctly it should do exactly what we do now but is laying the ground work for next steps?

That's the idea!

djmitche:
It looks like this would create the status when the first task is defined (which would occur as a result of the `queue.createTask` call that tc-github itself makes).  In fact, it would create the status once per task -- is it OK with Github to create the same status multiple times?

owlishDeveloper:
Oh - well, in case of Statuses API the status will get overwritten; that's a good point tho! Statuses API is for group status, so I need to have a slightly different logic there (it won't do to make many redundant requests). Good catch, @djmitche ! 🐟

owlishDeveloper:
I deployed on my cluster and tested. It works!

config.yml
```diff
@@ -3,6 +3,7 @@ defaults:
   app:
     jobQueue: null
     statusQueue: null
```
djmitche: I wonder if this should be "taskGroupQueue"?

schemas/v1/task-group-defined-message.yml
```diff
@@ -0,0 +1,28 @@
...
+  repository:
+    description: |
+      The GitHub `repository` which had an event.
+    type:         string
+    minLength:    {$const: github-identifier-min-length}
+    maxLength:    {$const: github-identifier-max-length}
+    pattern:      {$const: github-identifier-pattern}
+  taskGroupId:
+    type: string
+    description: The target URL to associate with this status. This URL will be linked from the GitHub UI to allow users to easily see the 'source' of the Status.
```
djmitche: If this is a URL, then `taskGroupId` is a confusing name for the field..

schemas/v1/task-group-defined-message.yml
```diff
@@ -0,0 +1,28 @@
...
+additionalProperties: false
+required:
+  - taskGroupId
```
djmitche, line 29: I would expect all three fields to be required?

schemas/v1/task-group-defined-message.yml
```diff
@@ -0,0 +1,28 @@
+$schema:  http://json-schema.org/draft-06/schema#
+title:        "Task Group Defined - Create Status"
+description: |
+  Create a commit status on GitHub.
+  Full specification on [GitHub docs](https://developer.github.com/v3/repos/statuses/#create-a-status)
```
djmitche, line 6: This description is from the perspective of the message receiver, but as an API definition, the service is the *sender* and the user is the receiver.  That is, as someone reading this description I can't decide to send a message to create a commit status.  My only option is to decide to listen for messages like this.  So, what would it mean for me to receive a message like this?

I think it's important not to suggest that this is the task-group equivalent of `queue.taskDefined`, since someone might otherwise get the wrong idea and depend on it for that purpose.

Come to think of it, one we drop status support, we will likely want to get rid of this message.  So perhaps the description should just say something like "Indicates that this service has created a new task group in response to a GitHub event.  This message is for internal use only and should not be relied on for other purposes."

src/handlers.js
```diff
@@ -389,25 +418,8 @@ async function jobHandler(message) {
...
-      owner: organization,
-      repo: repository,
-      sha,
-      state: groupState,
-      target_url,
-      description,
-      context: statusContext,
-    });
+
+    debug(`Trying to create status for ${organization}/${repository}@${sha} (${groupState}) in Azure`);
```
djmitche: Is this still accurate?

src/handlers.js
```diff
@@ -437,6 +449,53 @@ async function jobHandler(message) {
...
+  } catch (e) {
+    debug(`Creating tasks for ${organization}/${repository}@${sha} failed! Leaving comment on Github.`);
+    await this.createExceptionComment({instGithub, organization, repository, sha, error: e});
   }
 }
+
+/**
+ * If a task was defined, post the initial status to github
+ *
+ * @param message - taskDefined exchange message
```
djmitche: It looks like this is not taskDefined, but taskGroupDefined?

src/handlers.js
```diff
@@ -437,6 +449,53 @@ async function jobHandler(message) {
...
+ *   https://docs.taskcluster.net/docs/reference/platform/taskcluster-queue/references/events#taskdefined
+ * @returns {Promise<void>}
+ */
+async function taskGroupHandler(message) {
+  const {taskGroupId} = message.payload;
+
+  const debug = Debug(`${debugPrefix}:task-handler`);
+  debug(`Task was defined for ${taskGroupId}. Creating status...`);
```
djmitche: task group?

djmitche (commented):
So it seems like the main change here is to move status creation out of the webhook handler and into the Pulse message handlers.  I think the code looks fine, and I've made some comments around naming that would make this a bit clearer for the reader.

config.yml
owlishDeveloper: This queue will later be used for both Statuses and Checks. The former tracks group status, and the latter - well, it will depend. I was thinking of `resultStatus` <--- what do you think of that?

src/handlers.js
owlishDeveloper: I stuck "in Azure" at the back so it would be accurate. I'll change the wording so that it's clearer

djmitche: Sorry, I should have been more specific: instead of creating anything, it now sends a pulse message..

owlishDeveloper: Well, it creates a record in the db, that's what I wanted this log message to say

djmitche: Oh, I get it now :)  How about "Add .. to Builds table"?

owlishDeveloper: yeah, that's what I meant by "changing the wording to make it clearer" :)

djmitche (approved):
I like it!

PR #306, opened 2018-11-27
Bug 1507254 - Allow for some replication lag on github's part

imbstack:
This does the following:

*  adds a plugin to octokit to exponentially backoff requests that 404. The max it will wait with these hardcoded constants is ~12 seconds. We can make this configurable if we want as well. If this works for us for a while, perhaps we can upstream it for octokit/rest.js#900
* Fixes some tests that had fallen out of date with our actual logic
* Updates octokit to latest version

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -287,8 +287,8 @@ builder.declare({
   }
 
   // Not all webhook payloads include an e-mail for the user who triggered an event
-  let headUser = msg.details['event.head.user.id'].toString();
-  let userDetails = (await instGithub.users.getById({id: headUser})).data;
+  let headUser = msg.details['event.head.user.login'].toString();
+  let userDetails = (await instGithub.users.getByUsername({username: headUser})).data;
   msg.details['event.head.user.email'] = this.ajv.validate({type: 'string', format: 'email'}, userDetails.email)
     ? userDetails.email
     : msg.details['event.head.user.login'].replace(/\[bot\]$/, '') + '@users.noreply.github.com';
--- a/src/github-auth.js
+++ b/src/github-auth.js
@@ -1,7 +1,31 @@
 const debug = require('debug')('taskcluster-github:github-auth');
-const Github = require('@octokit/rest');
 const Promise = require('bluebird');
 const jwt = require('jsonwebtoken');
+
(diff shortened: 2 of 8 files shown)


owlishDeveloper (commented):
👍

src/api.js
```diff
@@ -287,8 +287,8 @@ builder.declare({
   }
 
   // Not all webhook payloads include an e-mail for the user who triggered an event
-  let headUser = msg.details['event.head.user.id'].toString();
-  let userDetails = (await instGithub.users.getById({id: headUser})).data;
```
djmitche, line 291: Is the `id` field no longer available?  For users `id` is permanent, but `username` can change, so using the `id` is preferred where possible.

src/github-auth.js
```diff
@@ -1,8 +1,32 @@
...
 
+const retryPlugin = (octokit, options) => {
+
+  const retries = 7;
+  const baseBackoff = 100;
+  const sleep = timeout => new Promise(resolve => setTimeout(resolve, timeout));
+
+  octokit.hook.wrap('request', async (request, options) => {
+    let response;
+    for (let attempt = 0; attempt < retries; attempt++) {
```
djmitche, line 13: I assume this form is used ironically, since the hipsters of last year hated it? :wink:

djmitche (commented):
I'm worried about `id` -> `username` as that might open the door to user impersonation..

src/api.js
imbstack: Apparently `getById` is undocumented? In [the release notes](https://github.com/octokit/rest.js/releases/tag/v16.0.0) they have a way to replace it with a generic request but they seem to think that the endpoint could go away at any time.

src/github-auth.js
imbstack: I honestly forgot which order the x;y;z went in! It's been that long since I wrote something like that.

Honestly it just worked better for my mental model of this.

src/api.js
djmitche: Hm, I guess GitHub isn't worried about this issue, then..

djmitche (approved):

owlishDeveloper (approved):

PR #307, opened 2018-11-28
Update more deprecated methods

imbstack:
This appears to work although while I was testing it out I found that the badges appear to not link to the correct build anymore? An example is the one on this project.


djmitche (approved):
Is there some way to mechanically compare the mock to the real thing, to check that the method names match?  Or, could we set up a "mock" and "real" with the real thing using a real (but test) app?

imbstack:
Theoretically there are tests for that in here already to do that.

https://github.com/taskcluster/taskcluster-github/blob/master/test/fake_github_test.js

PR #309, opened 2018-11-29
Status via Checks API

owlishDeveloper:


src/handlers.js
```diff
@@ -73,19 +106,30 @@ class Handlers {
...
+    ];
+
     // Listen for state changes to the taskcluster tasks and taskgroups
     // We only need to listen for failure and exception events on
     // tasks. We wait for the entire group to be resolved before checking
     // for success.
-    const statusBindings = [
-      queueEvents.taskFailed({schedulerId}),
+    const deprecatedStatusBindings = [
+      queueEvents.taskFailed({schedulerId}), // TODO: I think this is going to match EVERYTHING
```
owlishDeveloper: This does indeed match both the statuses and checks, so I will deal with that (this is not to ask for help, but to let you know I know this is a problem, and I'm on it)

schemas/v1/task-group-creation-requested.yml
```diff
@@ -24,6 +24,9 @@ properties:
   taskGroupId:
     type: string
     description: The id of the taskGroup that had been created.
+  reporting:
```
imbstack: Let's use an enum here as well

schemas/v1/taskcluster-github-config.v1.yml
```diff
@@ -19,6 +19,9 @@ properties:
       enum:
         - public
         - collaborators
+  reporting:
```
imbstack, line 22: Also add a description and type I believe, right?

src/handlers.js
```diff
@@ -10,6 +10,33 @@ const {consume} = require('taskcluster-lib-pulse');
...
+  neutral: 'It is neither good nor bad',
+  cancelled: 'Cancelled',
+  timed_out: 'Timed out',
+  action_required: 'Action required',
+  queued: 'Queued',
+  in_progress: 'In progress',
+  completed: 'Completed',
+};
+
+const CONCLUSIONS = { // maps queue exchange status to github checkrun conclusion
```
imbstack: I guess most of these aren't really exchange statuses but rather `reason`s for exception tasks. `completed`, `failed`, `exception` are queue statuses but the rest aren't.

src/handlers.js
```diff
@@ -10,6 +10,33 @@ const {consume} = require('taskcluster-lib-pulse');
...
+  completed: 'success', // TODO: remove this stupid rule from es-lint
+  failed: 'failure',
+  exception: 'failure',
+  'deadline-exceeded': 'timed_out',
+  canceled: 'cancelled',
+  superseded: 'neutral', // is not relevant anymore
+  'claim-expired': 'neutral', // will be retried
```
imbstack: I _think_ that you will only get a pulse message once the final retry has an exception with `claim-expired`. In that case it would be a failure here.

src/handlers.js
```diff
@@ -263,6 +327,80 @@ async function statusHandler(message) {
...
+        ...taskState,
+        owner: organization,
+        repo: repository,
+        check_run_id: checkRun.checkRunId, // TODO: a helpful output
+      });
+    } else {
+      const checkRun = await instGithub.checks.create({
+        owner: organization,
+        repo: repository,
+        name: `Task ${taskId}`, // <-- this needs to be unique 😩
```
imbstack: What if we fetch the task definition and use `task.metadata.name` here?

imbstack (commented):
Looking good so far!

schemas/v1/task-group-creation-requested.yml
owlishDeveloper: this prop is gone!

src/handlers.js
owlishDeveloper: yeah, forgot to edit the comment

src/handlers.js
owlishDeveloper: I had thought about it, and I don't like that option because I don't want to rely on users for having unique `task.metadata.name`. It's like "Hey, I can't solve this problem, so please be inconvenienced", and I doubt that's a realistic expectation anyways. People just copy-paste metadata generally. What will happen is this: then they will see one checkrun for one task, while their YAML has 5. Instead of reading the docs and editing the YML, they will file a bug in tc-gh. (and that will be fair!)

I plan to think about this some more. It's not to ask for help, more like a `TODO`. However, I don't want this to be a blocker for the first pass either.

owlishDeveloper: I mean, even if I implement a check for `task.metadata.name` uniqueness in `tc-yml`... I wonder if that would make creating tc-yml more complicated and obnoxious... 
> "Hey, I can't solve this problem, so please be inconvenienced"

src/handlers.js
owlishDeveloper: I thought that would mean a retry - i.e. a new task is created? (I don't know how queue is working here, tho. So I'll trust you, I think.... So, is that true for the rest of `neutral`s here? Should I just replace that with `failure`?)

imbstack: It's a bit confusing -- `retry` vs `rerun`. A rerun happens within a task and a retry creates a new task. Exceptions result in reruns. I might've even said it results in a retry at some point when I was being imprecise so the confusion is easy to create.

Yeah, let's call it a failure.

src/tc-yaml.js
```diff
@@ -252,19 +252,24 @@ class VersionOne extends TcYaml {
...
-          taskId: task.taskId,
-          task: _.omit(
-            _.extend(task, {schedulerId: cfg.taskcluster.schedulerId}),
-            'taskId'),
+        const {taskId, ...taskWithoutTaskId} = task;
+        const b = {
+          taskId,
+          task: {
+            ...taskWithoutTaskId,
+            schedulerId: cfg.taskcluster.schedulerId,
```
imbstack, line 273: Why not add `schedulerId` first and then just return `{taskId, ...taskWithoutTaskId}`?

src/tc-yaml.js
```diff
@@ -252,19 +252,24 @@ class VersionOne extends TcYaml {
       }
 
       config.tasks = config.tasks.map(task => {
-        task = _.defaults(task, {
+        task = Object.assign({
           taskId: defaultTaskId,
           taskGroupId: defaultTaskGroupId,
           created: now,
-        });
+          routes: config.reporting ? [cfg.app.checkTaskRoute] : [cfg.app.statusTaskRoute],
```
imbstack: What happens here if the user specifies routes on their task directly? Do these get overwritten?

I guess we'll have to say in our docs that you need to add specific routes to make up for the lost defaults if you have any yourself?

imbstack: Forgive me if this was already discussed.

src/handlers.js
```diff
@@ -10,6 +10,33 @@ const {consume} = require('taskcluster-lib-pulse');
...
+
+const CONCLUSIONS = { // maps status communicated bu the queue service to github checkrun conclusions
+  completed: 'success', // TODO: remove this stupid rule from es-lint
+  failed: 'failure',
+  exception: 'failure',
+  'deadline-exceeded': 'timed_out',
+  canceled: 'cancelled',
+  superseded: 'neutral', // is not relevant anymore
+  'claim-expired': 'failure',
+  'worker-shutdown': 'neutral', // will be retried
```
imbstack: I think this and `intermittent-task` are also failures if they happen on the last run attempt. @djmitche might be able to correct me though.

src/handlers.js
```diff
@@ -266,6 +340,82 @@ async function statusHandler(message) {
...
+  });
+
+  let {organization, repository, sha, eventId, eventType, installationId} = build;
+
+  let debug = Debug(`${debugPrefix}:${eventId}`);
+  debug(`Handling state change for task ${taskId} in group ${taskGroupId}`);
+
+  let taskState = {
+    status: 'completed',
+    conclusion: CONCLUSIONS[reasonResolved || state],
```
imbstack: Maybe we need to have a case here for if neither are already defined? Default to failure I expect. This could be the case if we add a new `reason` to exceptions and don't update here for instance. Another option would be to throw an error on our end so that we fix it.

imbstack (approved):
This is looking really good. Maybe a test deployment soon?

src/handlers.js
djmitche: I'm comfortable continuing to call it neutral even in that case -- it wasn't the code's fault.

src/tc-yaml.js
owlishDeveloper: > What happens here if the user specifies routes on their task directly? Do these get overwritten?

I would hope so! Otherwise the routing won't work, and the functionality will be broken.

Do we want to allow specifying routes? And if so - why? Do people currently specify routes? If so - why, without the bindings?

My understanding of the purpose of routes was that we need them to separate Checks from Statuses, or to have the pulse messages come through two separate channels. Once Statuses are deprecated and gone from the codebase, routes (and the bindings) will be gone as well (?)

This looks like a design-stage topic to me. I don't recall us discussing this during the rfc stage. I'm not against implementing that if we decide it's needed, but not in this PR.

src/tc-yaml.js
owlishDeveloper: I'm not sure I see the difference between the two options

src/handlers.js
owlishDeveloper: state is a required prop on these exchange messages

owlishDeveloper: oh, you mean if there's a new element in the `enum`. Yes, that's a great idea! Why default to failure tho? What if that is not a failure type of reason, but somebody PR would be blocked? Maybe `Action required`? As in, "hey, we got an unknown message please click the link and check it out". Or `neutral`, with the reason in the description of the checkrun

src/tc-yaml.js
owlishDeveloper: oh, `const b` is a debugging leftover! I'll clean it up, it shouldn't be there. Thank you for noticing!

src/tc-yaml.js
owlishDeveloper: ...unless you mean people would add routes to have something else listening to those. If that's the case - I can organize pushing into an array here. Let me know if that's what you meant (and I'm still curious about the use cases)

owlishDeveloper: (I'll just go ahead and make it so that it pushes. It seems to me you meant a third application listening to those - and that's a good point. I'm in such a panicky state today)

src/handlers.js
imbstack: Sure, both sound good to me. Also log something that we can alert on later!

src/tc-yaml.js
imbstack: Yeah, people add routes for things like [notifications](https://github.com/taskcluster/taskcluster-notify#taskcluster-notifications-service). I think pushing if something is already there makes sense! Unfortunately we have a limit (I think it is 10) to the number of routes you can add. Maybe there needs to be some sort of check that we won't go over that? For now just push and we can figure that out later.

src/handlers.js
owlishDeveloper: oh, log something - good point!

src/tc-yaml.js
owlishDeveloper: Oh that's actually awesome! I didn't realize we had that functionality. If there's a limit, something must have the check somewhere. Imma check it out

owlishDeveloper: [Found it](https://github.com/taskcluster/taskcluster-queue/blob/3b2960adf20ccb960a474fd68a526b4c389868b4/schemas/v1/task.yml#L94) - I would hope if people have 68 routes in their yml it would fail the schema check? 

UPD: [Checked that](https://github.com/owlishDeveloper/helloWorld/pull/93/files) - the schema checks happily passes

I would leave this for now (64 seems like a large enough number), but in a separate PR I would like to enforce task definition schema checks

owlishDeveloper: UPD2: [The schema check failed after all!](https://github.com/owlishDeveloper/helloWorld/pull/93) Yay. 

In a separate PR then I will deal with the bug of creating status when tasks were not submitted. This is not the first time I see it

owlishDeveloper:
Thank you for reviewing, bstack! How about a test deployment tomorrow?

src/tc-yaml.js
owlishDeveloper: Added mention to the documentation for now.

src/tc-yaml.js
```diff
@@ -252,18 +252,24 @@ class VersionOne extends TcYaml {
...
+        task = Object.assign({
           taskId: defaultTaskId,
           taskGroupId: defaultTaskGroupId,
           created: now,
-        });
+        }, task);
         defaultTaskId = slugid.nice(); // invent a new taskId for the next task
 
+        const {taskId, ...taskWithoutTaskId} = task;
+        console.log(task.routes);
```
djmitche: debug console.log

src/handlers.js
```diff
@@ -10,6 +10,34 @@ const {consume} = require('taskcluster-lib-pulse');
...
+  neutral: 'It is neither good nor bad',
+  cancelled: 'Cancelled',
+  timed_out: 'Timed out',
+  action_required: 'Action required',
+  queued: 'Queued',
+  in_progress: 'In progress',
+  completed: 'Completed',
+};
+
+const CONCLUSIONS = { // maps status communicated bu the queue service to github checkrun conclusions
```
djmitche: "by"

src/handlers.js
```diff
@@ -68,22 +107,32 @@ class Handlers {
       githubEvents.release(),
     ];
 
-    const schedulerId = this.context.cfg.taskcluster.schedulerId;
     const queueEvents = new taskcluster.QueueEvents({rootUrl: this.rootUrl});
 
+    const statusBindings = [
+      queueEvents.taskFailed(`route.${this.context.cfg.app.checkTaskRoute}`),
+      queueEvents.taskException(`route.${this.context.cfg.app.checkTaskRoute}`),
+      queueEvents.taskCompleted(`route.${this.context.cfg.app.checkTaskRoute}`),
```
djmitche, line 112: I don't see this config in `config.yml`.

src/handlers.js
```diff
@@ -68,22 +107,32 @@ class Handlers {
...
-      queueEvents.taskException({schedulerId}),
-      queueEvents.taskGroupResolved({schedulerId}),
+    const deprecatedStatusBindings = [
+      queueEvents.taskFailed(`route.${this.context.cfg.app.statusTaskRoute}`),
+      queueEvents.taskException(`route.${this.context.cfg.app.statusTaskRoute}`),
+      queueEvents.taskGroupResolved(`route.${this.context.cfg.app.statusTaskRoute}`),
+    ];
+
+    // Listen for taskGroupCreationRequested event to create initial status on github
+    const deprecatedBindings = [
```
djmitche: deprecatedInitialStatusBindings maybe?

src/handlers.js
```diff
@@ -68,22 +107,32 @@ class Handlers {
...
+
     // Listen for state changes to the taskcluster tasks and taskgroups
     // We only need to listen for failure and exception events on
     // tasks. We wait for the entire group to be resolved before checking
     // for success.
-    const statusBindings = [
-      queueEvents.taskFailed({schedulerId}),
-      queueEvents.taskException({schedulerId}),
-      queueEvents.taskGroupResolved({schedulerId}),
+    const deprecatedStatusBindings = [
```
djmitche: deprecatedResultStatusBindings?

src/handlers.js
```diff
@@ -328,10 +493,13 @@ async function jobHandler(message) {
     throw e;
   }
 
+  let reporting;
   // Check if this is meant to be built by tc-github at all.
   // This is a bit of a hack, but is needed for bug 1274077 for now
   try {
     let c = yaml.safeLoad(repoconf);
+    reporting = c.reporting;
+    console.log(reporting);
```
djmitche: debugging console.log

src/handlers.js
```diff
@@ -417,72 +585,82 @@ async function jobHandler(message) {
     }
   }
 
-  try {
-    taskGroupId = graphConfig.tasks[0].task.taskGroupId;
-    debug(`Creating tasks for ${organization}/${repository}@${sha} (taskGroupId: ${taskGroupId})`);
-    await this.createTasks({scopes: graphConfig.scopes, tasks: graphConfig.tasks});
+  taskGroupId = graphConfig.tasks[0].task.taskGroupId;
```
djmitche, line 578: what if the tasks have different taskIds?

test/github-auth.js
```diff
@@ -107,6 +107,26 @@ class FakeGithub {
           throwError(404);
         }
       },
+      'checks.create': async ({owner, repo, name, head_sha, output, details_url, actions, status, conclusion}) => {
+        if (repo === 'no-permission') {
+          throwError(403);
+        }
+
+        const check_run_id = Math.floor(Math.random()*(9999-1000));
```
djmitche: is this missing a +1000?

src/handlers.js
```diff
@@ -417,72 +585,82 @@ async function jobHandler(message) {
...
-      assert.equal(build.eventId, message.payload.eventId);
-    });
+    }
 
-    debug(`Publishing status exchange for ${organization}/${repository}@${sha} (${groupState})`);
-    await context.publisher.taskGroupDefined({taskGroupId, organization, repository});
+  }).catch(async e => {
+    return debug(`Failed to publish to taskGroupCreationRequested exchange 
+    for ${organization}/${repository}@${sha} with the error: ${JSON.stringify(e, null, 2)}`);
+  });
```
djmitche: Why is this new code in the pre-async/await `.then`-style?

src/tc-yaml.js
```diff
@@ -252,18 +252,24 @@ class VersionOne extends TcYaml {
...
         return {
-          taskId: task.taskId,
-          task: _.omit(
-            _.extend(task, {schedulerId: cfg.taskcluster.schedulerId}),
-            'taskId'),
+          taskId,
+          task: {
+            ...taskWithoutTaskId,
+            schedulerId: cfg.taskcluster.schedulerId,
```
djmitche, line 273: Does this mean that all github tasks must still have the same schedulerId?

djmitche (approved):

src/handlers.js
owlishDeveloper: It's just one of those mysteries. Like, why do headphone wires tangle up while the headphones are lying on the table? 😂 

(I used `then` to make absolutely sure things happen in the order I expect them to. I was seeing weird behavior, and concluded that two separate `await`s work in parallel, and not in series as I had assumed. However, now that I know RabbitMQ persists its state so valiantly, I suspect that was the reason behind the behavior I observed. Besides, now that the flow has changed 10 commits ago, there's no reason for these anymore - I will be cleaning up here)

src/tc-yaml.js
owlishDeveloper: (I will be removing this. I wanted to see the output while running tests with no `DEBUG` set)

(I am so sorry for all this 40+ commit mess. I'm cleaning this up!)

src/handlers.js
owlishDeveloper: this part is gone now

src/handlers.js
owlishDeveloper: taskIds?

test/github-auth.js
owlishDeveloper: I think I just wanted a random 4-digit number for some reason 🤔 I don't think tc-gh is making any decision based on check_run_id, so the values of this variable are probably inconsequential?

src/tc-yaml.js
owlishDeveloper: I don't think they _must_. This code doesn't care about scheduler id because we have  more precise filtering. However, I think having it in the task definition won't hurt, and might be useful. You think it's better not to have it anymore?

src/handlers.js
djmitche: Right, sorry, I meant what if they have different taskGroupIds -- like maybe my .taskcluster.yml creates a taskGroup for all of the test tasks, and a separate task group that represents a release pipeline?

test/github-auth.js
djmitche: That's what I figured, but it's really just a random number from zero to 8999, but adding 1000 to it would get you 1000-9999 which is the "four digits" you wanted, right?  This totally doesn't matter to the review, I just noticed it and asked..

src/tc-yaml.js
djmitche: As written, any schedulerId specified in the .taskcluster.yml is overridden, so users can't set it.  But we could change that later.  The important thing in this PR is that the schedulerId is no longer used to filter, as you've said.

src/exchanges.js
```diff
@@ -116,17 +123,19 @@ exchanges.declare({
...
-  name:               'taskGroupDefined',
-  title:              'GitHub release Event',
+  exchange:           'task-group-creation-requested',
+  name:               'taskGroupCreationRequested',
+  title:              'tc-gh requested the Queue service to create all the tasks in a group',
   description: [
-    'used for creating status indicators in GitHub UI using Statuses API',
+    'supposed to signal that taskCreate API has been called for every task in the task group',
+    'for this particular repo and this particular organization',
+    'currently used for creating initial status indicators in GitHub UI using Statuses API',
```
djmitche: If I'm understanding correctly, then, messages on this exchange have a primary routing key similar to the queue's exchanges, and a CC'd routing key containing either `route.checks` or `route.statuses`, and that's what the listeners use to distinguish which handler should handle a particular message.  Similar CC routes are used for messages from the queue's exchanges, by setting `task.routes`.

It might be worth including some information about that in the description here, to help the next person looking at the code :)

src/handlers.js
owlishDeveloper: ah, I see. Yes, good point

test/github-auth.js
owlishDeveloper: aaaah, haha, yes! Does seem to be a silly error. I'll add 1000

src/tc-yaml.js
owlishDeveloper: Yeah, I think we used to override it because we used it for filtering before. I decided not to change this at this point

src/handlers.js
owlishDeveloper: I am leaving this as it is for now. I would like to tackle this in a separate follow-up PR, because this is something that I should have done when I added v1.

owlishDeveloper:
This work continues in #323

PR #323, opened 2019-01-03
Add support for github checks API

owlishDeveloper:


djmitche (approved):

djmitche:
I confirmed this pretty much matches #309 so no need for @imbstack to look.

